PR #12, opened 2020-08-18
Force to close the database connection after each task

### LINKED ISSUE #11: Celery workers are unable to reinitialize a lost database connection
An `InterfaceError: connection already closed` error occurs regularly, rendering the celery worker unusable.

See https://github.com/celery/django-celery/issues/121 and https://stackoverflow.com/questions/31504591/interfaceerror-connection-already-closed-using-django-celery-scrapy

aperrin66:
Resolves #11

### DIFF
--- a/geospaas_processing/tasks.py
+++ b/geospaas_processing/tasks.py
@@ -2,23 +2,33 @@
 import errno
 import os
 
-from celery import Celery
-from celery.utils.log import get_task_logger
+import celery
+from django.db import connection
 
 import geospaas_processing.utils as utils
 from .converters import IDFConverter
 from .downloaders import DownloadManager, TooManyDownloadsError
 
 
-LOGGER = get_task_logger(__name__)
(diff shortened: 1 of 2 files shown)


mlissner:
I'm curious: Can anybody say if this worked? I'm probably going to implement it, but we have hundreds of tasks, so it will be a pain. If anybody knows that this worked, that's be great info.

aperrin66:
@mlissner It did not work at the time. I encountered the error again after putting this into place.
I have since updated to more recent versions of Celery (5.2) and Django (3.2).
I have not encountered the problem since then, although I never managed to properly replicate the issue, so I can't guarantee that it is fixed.

mlissner:
Thank you, that's really great info. Oh, Celery...

aperrin66:
@mlissner In case this is of interest to you, I ran into this problem again (I did not change Celery or Django versions).

PR #13, opened 2020-09-08
FTP synchronization

aperrin66:
Add a task that publishes the result files to an FTP server accessible via SCP.

There is also a small refactoring of disk space management utilities.

FYI @aanersc


aperrin66:
I added a few additional changes:
- initiate the README
- add the ability to log to Graylog

akorosov (approved):
Nice usage of decorator for locking!
And the Storage classes are are also cool!

PR #16, opened 2020-10-15
#15 default  value file name is entry_id field of dataset

### LINKED ISSUE #15: The default value of filename should be returned by the server either from header, or from url instead of an empty one
Downloaded dataset should be named based on its entry_id for all cases.

For copernicus harvested file the header of the response of the download link provide the associated part named `'filename=' `for find the name of the dataset based on it. But for the osisaf and other future download links, the header of the response whould be something like:
image

So no information can be gained about the filename from such a header. Thus, this line
https://github.com/nansencenter/django-geo-spaas-processing/blob/7793104419a786111342e121fbaae111acb72d64/geospaas_processing/downloaders.py#L65
should be corrected for solving this issue.

aanersc:
closes #15 
@aperrin66 
@akorosov 

Simple modification for solving the issue of name described in #15

### DIFF
--- a/geospaas_processing/downloaders.py
+++ b/geospaas_processing/downloaders.py
@@ -47,7 +47,7 @@ class HTTPDownloader(Downloader):
     CHUNK_SIZE = 1024 * 1024
 
     @staticmethod
-    def extract_file_name(response,url):
+    def extract_file_name(response, url):
         """Extracts the file name from the Content-Disposition header of an HTTP response"""
         filename_key = 'filename='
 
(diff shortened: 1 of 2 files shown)


aanersc:
@aperrin66 
@akorosov 
I forgot to change the tests based on the demand of this PR. Now, it is ready to be merged. You can review it now.

aperrin66 (changes requested):
These modifications should not be made in the `extract_file_name()` method.
As indicated in its docstring, the role of this method is to extract the file name **from the Content-Disposition header of an HTTP response**.
Please keep in mind that, as much as possible, a function or method should do only one thing.

I am also not comfortable using the `entry_id` field to get the file name. This field does not necessarily contain the file name. It contains a unique ID, which happens to be the file name in most cases, but it is not a guaranteed behavior.

PR #20, opened 2020-10-26
Issue19 new script for multi folder

aanersc:
@aperrin66 
@akorosov 

New dockerfile and new script is added to this repo as a generic code. These files use another config file that stored in a folder different from the original config file of repo. The new folder name is `multi_folder_config`. Everying is done during the building the image and there is no need to mount the files for downloading. We only need to mount to see the result of downloading (files) from outside the container.


aanersc:
@akorosov 
@aperrin66 
The new entry point is based on the new slim image! So a tiny thing should be corrected in the dockerfile of it. it is done in f21dbd4 and now it is ready for review.

Dockerfile_multi_folder
```diff
@@ -0,0 +1,14 @@
+ARG BASE_IMAGE=nansencenter/geospaas:2.0.0-slim
+FROM ${BASE_IMAGE}
+RUN pip install --no-cache-dir django-celery-results==1.2
+WORKDIR /tmp/setup
+COPY setup.py README.md ./
+COPY geospaas_processing ./geospaas_processing
+RUN python setup.py bdist_wheel && \
+    pip install -v dist/geospaas_processing-*.whl && \
+    cd /tmp && rm -rf /tmp/setup/
+COPY multi_folder_download.py /multi_folder_download.py
```
akorosov, line 10: That script should rather become part of the package geospaas_processing and be installed automatically by setup.py

multi_folder_download.py
```diff
@@ -0,0 +1,56 @@
...
+from dateutil.tz import tzutc
+from django.contrib.gis.geos import GEOSGeometry
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'geospaas_processing.settings')
+
+import geospaas_processing.downloaders as downloaders
+from geospaas.catalog.models import Dataset
+
+
+def main_function(ar):
```
akorosov, line 20: Change name.

multi_folder_download.py
```diff
@@ -0,0 +1,56 @@
...
+    download_manager = downloaders.DownloadManager(
+        download_directory=ar.mount_dir.rstrip(os.path.sep)+os.path.sep+ar.dir_struct.lstrip(
+            os.path.sep),
+        use_file_prefix=strtobool(ar.use_filename_prefix),
+        provider_settings_path=os.path.join(os.path.dirname(
+            __file__), ar.config_folder, 'provider_settings.yml'),
+        time_coverage_start__gte=designated_start_time,
+        time_coverage_end__lte=designated_end_time,
+        geographic_location__geometry__intersects=GEOSGeometry(ar.geometry),
+        dataseturi__uri__contains=ar.criteria)
```
akorosov, line 39: Make it generic and allow search not only by `dataseturi__uri`. For example if I want to search by source I want to call this script with option `-criterium source__instrument__short_name__icontains='SAR'`
We've discussed this how to implement that. criteria can be a dictionary that you generate from several `-c` options and then pass to DownloadManager as `**kwargs`

multi_folder_download.py
```diff
@@ -0,0 +1,56 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description='Process the arguments of entry_point (all must be in str)')
+    parser.add_argument('-md', '--mount_dir', required=True, type=str)
+    parser.add_argument('-cr', '--criteria', required=True, type=str)
+    parser.add_argument('-st', '--start_time', required=True, type=str)
+    parser.add_argument('-et', '--end_time', required=True, type=str)
+    parser.add_argument('-rtf', '--rel_time_flag', required=True, type=str)
+    parser.add_argument('-rtl', '--rel_time_lag', required=True, type=str)
```
akorosov, line 50: I think it is better to use just one begin date instaed of two:
either
`multi_folder_download.py -begin 2020-10-10 -end 2020-10-20 ... -other options`
or
`multi_folder_download.py -begin -10 --relative-time ... -other options`

So remove  `rel_time_lag` and convert `-rtf` to boolean (see action='store_true')

multi_folder_download.py
```diff
@@ -0,0 +1,56 @@
...
+        time_coverage_start__gte=designated_start_time,
+        time_coverage_end__lte=designated_end_time,
+        geographic_location__geometry__intersects=GEOSGeometry(ar.geometry),
+        dataseturi__uri__contains=ar.criteria)
+    download_manager.download()
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description='Process the arguments of entry_point (all must be in str)')
+    parser.add_argument('-md', '--mount_dir', required=True, type=str)
```
akorosov, line 45: THis is not needed here but in the master script which will actually mount the dirs.

akorosov (changes requested):

multi_folder_download.py
aanersc: Change name of which one? `main_function` or `ar`. I name it as ar since it is repeated many times in the function and it is better not to slipt the lines several times!

If you mean main_funciton, I think it is a good name for it because this is the main function that is called.

multi_folder_download.py
aanersc: This is needed here for the source code of downloading to find where to put the downloaded files. It is used in this line and it must be used like that:
https://github.com/nansencenter/django-geo-spaas-processing/blob/f21dbd4887a22328e74b4f76fb049ccab56f0891/multi_folder_download.py#L31

multi_folder_download.py
aanersc: As I have mentioned, since we use `dockerpy` we can not convey another type different from `str`. So, action='store_true' can't be implemented.

I think it is good to discriminate the input values from the very beginning in order to avoid confusion for the user. start_time should receive the date of start and rel_time_lag should receive hours! Since these two are two different things, it is better to receive them as two different arguments. What is your opinion @aperrin66 ?

Dockerfile_multi_folder
aperrin66: I agree, and to take it a bit further, we could make a "cli" sub-package containing several modules allowing access the functionalities of geospaas_processing from a command line.

`multi_folder_download.py` can be simply renamed to `download.py`.

We can have a structure like:
```
geospaas_processing/
    cli/
        download.py
        convert.py
    ...
```

Dockerfile_multi_folder
```diff
@@ -0,0 +1,14 @@
+ARG BASE_IMAGE=nansencenter/geospaas:2.0.0-slim
```
aperrin66, line 1: Comment on the whole file: a few empty lines to separate the sections would be nice.

Dockerfile_multi_folder
```diff
@@ -0,0 +1,14 @@
...
+FROM ${BASE_IMAGE}
+RUN pip install --no-cache-dir django-celery-results==1.2
+WORKDIR /tmp/setup
+COPY setup.py README.md ./
+COPY geospaas_processing ./geospaas_processing
+RUN python setup.py bdist_wheel && \
+    pip install -v dist/geospaas_processing-*.whl && \
```
aperrin66, line 8: That can can probably be replaced by `python setup.py install` (needs testing).
I don't remember why I built the wheel instead of installing directly in the other Dockerfile.

multi_folder_config/provider_settings.yml
```diff
@@ -0,0 +1,10 @@
+---
```
aperrin66, line 1: This is ADAS specific configuration, it should not be in this repository. 
This is typically the kind of thing which has its place in the ansible repository, and can then be mounted in the container.

Dockerfile_multi_folder
aperrin66, line 1: The name of this file give no idea of what it's for.
If we go with the CLI thing, this dockerfile could be named Dockerfile_cli, and the other one Dockerfile_worker.

multi_folder_download.py
aperrin66: When writing script which are going to be called from the command line, please keep to this simple convention:
```python
def main():
    print("Hello World!")  # the code goes here

if __name__ == "__main__":
    main()
```
See https://realpython.com/python-main-function/ for more details.

multi_folder_download.py
```diff
@@ -0,0 +1,56 @@
+import argparse
```
aperrin66, line 1: Please add a docstring for this script

multi_folder_download.py
aperrin66: Then it should not be called "mount_dir" but something like "download_dir".

aperrin66, line 45: About the arguments parsing:
- favor one-letter short options
- add a `help` string to each `add_argument()` call
- be coherent between the short and long name of an option. For example, in `parser.add_argument('-cdir', '--config_folder'...`, either use "folder" or "directory", but not one of each.

multi_folder_download.py
```diff
@@ -0,0 +1,56 @@
...
+def main_function(ar):
+    if strtobool(ar.rel_time_flag):
+        designated_start_time = datetime.now().replace(tzinfo=tzutc())+ relativedelta(
+            hours=-int(ar.rel_time_lag))
+        designated_end_time = datetime.now().replace(tzinfo=tzutc())
+    else:
+        designated_start_time = datetime.strptime(ar.start_time, "%Y.%m.%d:%H").replace(
+            tzinfo=tzutc())
+        designated_end_time = datetime.strptime(ar.end_time, "%Y.%m.%d:%H").replace(tzinfo=tzutc())
+    downloaders.DownloadManager.MAX_DOWNLOADS = 500
```
aperrin66, line 29: This is now an argument in the constructor of `DownloadManager`.

multi_folder_download.py
aperrin66: I agree with Anton. I find it more confusing to have two options for the same thing than having a flag modifying the behavior of an option. How can the user know what will happen if they specify both options in one call? With Anton's solution there is no ambiguity.

Also, I don't see why you couldn't use `action='store_true'`.

aanersc: @aperrin66 
@akorosov 
we use `dockerpy` somewhere in the deploy repo. So we need to pass all the arguments as a string! because we have to pass them as a list of strings to the dockerpy. This way, dockerpy can call the module with all its arguments. Otherwise it does not work and return the timeout error after 60 seconds. Because it cant parse the arguments that are sent to `client.containers.run()` command

akorosov: @aanersc , did you check how to use `action='store_true'` ?

It allows to have options like:

```python
my_script.py --foo

# in the script:
parser.add_argument('--foo', action='store_true')

args = parser.parse()
if args.foo:
    # do somtheing when script is called with --foo
else:
    # do somtheing when script is called without --foo
```

multi_folder_download.py
akorosov: The download directory is already there: `--dir_struct`

multi_folder_download.py
aanersc: Yeah, I know! :)
You will see in the next commit.

multi_folder_download.py
aanersc: Done in the next commit.

Dockerfile_multi_folder
aanersc: Done in the next commit.

multi_folder_download.py
aanersc: Done in the next commit

multi_folder_download.py
aanersc: DOne in the next commit.

multi_folder_download.py
aanersc: Done in the next commit.

multi_folder_download.py
aanersc: Done in the next commit.

Dockerfile_multi_folder
aanersc: Done in the next commit.

Dockerfile_multi_folder
aanersc: Done in the next commit.

multi_folder_config/provider_settings.yml
aanersc: Done in the next commit.

multi_folder_download.py
aanersc: Done in the next commit.

multi_folder_download.py
aanersc: Done in the next commit

multi_folder_download.py
aanersc: Done in the next commit

Dockerfile_multi_folder
aanersc: Done in the next commit.

aanersc:
@akorosov 
@aperrin66 
Everything is addressed in the recent commits. It is working with a file file that will be pushed tomorrow to the other repo(adas).

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    help="directory_structure is must be acceptable by python 'strftime'")
+    parser.add_argument('-p', '--use_filename_prefix', action='store_true',
+    help="The flag that distinguishes between the two cases of having files WITH or WITHOUT file "
+    "prefix when downloaded")
+    parser.add_argument('-g', '--geometry', required=False, type=str,
+    help="The 'wkt' string of geometry which is acceptable by 'GEOSGeometry' of django")
+    parser.add_argument('-c', '--config_file', required=False, type=str,
+    help="The absolute path to the config file that is needed for configuring the downloading "
+    "process. default is the same place of the 'download.py' file")
+    parser.add_argument('-l', '--limitation', required=True, type=str,
```
akorosov, line 80: I would rather use `append` functionality:
```python
parser.add_argument('-q','--query', action='append', help='<Required> query look-up', required=True)
# Use like:
# python download.py -q dataseturi__uri__contains=\'osisaf\' -q source__instrument__short_name__icontains=\'AMSR2\'
```

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+        time_coverage_start__gte=designated_begin,
+        time_coverage_end__lte=designated_end,
+        geographic_location__geometry__intersects=GEOSGeometry(ar.geometry),
+        **cumulative_limitation
+    )
+    download_manager.download()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
akorosov, line 55: Create a separate function for parsing arguments. Then it will be easier for testing.

akorosov (changes requested):
Please rebase on master first. It is hard to see what is coming from where.

Do you have unit tests for the script?

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    cumulative_limitation = {}
+    limitation_list = json.loads(ar.limitation)
+    for limitation in limitation_list:
+        cumulative_limitation.update(limitation)
+    if ar.rel_time_flag:
+        designated_begin = datetime.now().replace(tzinfo=tzutc()) + relativedelta(
+            hours=-abs(int(ar.begin)))
+        designated_end = datetime.now().replace(tzinfo=tzutc())
+    else:
+        designated_begin = datetime.strptime(ar.begin, "%Y.%m.%d:%H").replace(
```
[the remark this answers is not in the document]
aanersc: @akorosov 
I have used `%Y.%m.%d:%H` because the user may want to download the data from a specific hour of a day!
DO you mean only just by searching the days is enough? (by using `%Y-%m-%d`).
If we use days only, then we can not execute the code every 12 hours or every 8 hours! Because the code can distinguish between the hours! We can only run it once a day.
I think this cli is supposed to work based on hours in addition to days. Minutes and seconds are too much details and they may just become boring when working with them. What do you decide @akorosov ?

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+            tzinfo=tzutc())
+        designated_end = datetime.strptime(ar.end, "%Y.%m.%d:%H").replace(tzinfo=tzutc())
+    download_manager = downloaders.DownloadManager(
+        download_directory=ar.down_dir.rstrip(os.path.sep)+os.path.sep+ar.dir_struct.lstrip(
+            os.path.sep),
```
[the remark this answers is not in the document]
aanersc: @akorosov 
1.
The mounting action needs to have a specific folder. Since we want to see and access what we have downloaded from outside the container, we need to mount some specific folder. Thus, we need to break the address into different parts in order to make them visible from outside the container. That is the reason.
2. Done in the next commit.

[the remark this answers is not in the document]
aperrin66: Docker issues should not influence this script.
If there is a need for separating the path, it must be dealt with in the code which runs the container.

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_argument('-d', '--down_dir', required=True, type=str,
+    help="This must be an absolute path which downloaded files store in it")
+    parser.add_argument('-b', '--begin', required=True, type=str,
+    help="This must be an a time (starting point of the time) in the format of '%Y.%m.%d:%H' which "
+    "is acceptable by python strptime or the extend of time lag from now in the case of relative "
+    "time flag (-r) in the terms of hours")
+    parser.add_argument('-e', '--end', required=True, type=str,
+    help="This must be an a time (ending point of the time) in the format of '%Y.%m.%d:%H' which "
+    "is acceptable by python strptime. It is an effective option only in the case of ABSENCE of "
+    "relative time flag (-r).")
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description='Process the arguments of entry_point (all must be in str)')
+    parser.add_argument('-d', '--down_dir', required=True, type=str,
+    help="This must be an absolute path which downloaded files store in it")
+    parser.add_argument('-b', '--begin', required=True, type=str,
+    help="This must be an a time (starting point of the time) in the format of '%Y.%m.%d:%H' which "
+    "is acceptable by python strptime or the extend of time lag from now in the case of relative "
+    "time flag (-r) in the terms of hours")
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_argument('-e', '--end', required=True, type=str,
+    help="This must be an a time (ending point of the time) in the format of '%Y.%m.%d:%H' which "
+    "is acceptable by python strptime. It is an effective option only in the case of ABSENCE of "
+    "relative time flag (-r).")
+    parser.add_argument('-r', '--rel_time_flag', required=False, action='store_true',
+    help="The flag that distinguishes between the two cases of time calcation (1.time-lag from now "
+    "2.Two different points in time) based on its ABSENCE or PRESENCE in the arguments.")
+    parser.add_argument('-s', '--dir_struct', required=False, type=str,
+    help="directory_structure is must be acceptable by python 'strftime'")
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+        **cumulative_limitation
+    )
+    download_manager.download()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description='Process the arguments of entry_point (all must be in str)')
+    parser.add_argument('-d', '--down_dir', required=True, type=str,
+    help="This must be an absolute path which downloaded files store in it")
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    for limitation in limitation_list:
+        cumulative_limitation.update(limitation)
+    if ar.rel_time_flag:
+        designated_begin = datetime.now().replace(tzinfo=tzutc()) + relativedelta(
+            hours=-abs(int(ar.begin)))
+        designated_end = datetime.now().replace(tzinfo=tzutc())
+    else:
+        designated_begin = datetime.strptime(ar.begin, "%Y.%m.%d:%H").replace(
+            tzinfo=tzutc())
+        designated_end = datetime.strptime(ar.end, "%Y.%m.%d:%H").replace(tzinfo=tzutc())
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    download_manager = downloaders.DownloadManager(
+        download_directory=ar.down_dir.rstrip(os.path.sep)+os.path.sep+ar.dir_struct.lstrip(
+            os.path.sep),
+        provider_settings_path=ar.config_file,
+        # 400 dataset per day is allowed
+        max_downloads=400*(abs((designated_end-designated_begin).days)+1),
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
aanersc: Discussed! Done in the next commit

geospaas_processing/cli/download.py
aanersc: Done in the next commit

aanersc:
@akorosov 
@aperrin66 
All addressed in 1918044.

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+from geospaas.catalog.models import Dataset
+import geospaas_processing.downloaders as downloaders
+
+
+def main(arg):
+    cumulative_query = {}
+    if arg.query:
+        query_list = json.loads(arg.query)
+        for query in query_list:
+            cumulative_query.update(query)
```
akorosov, line 32: This is inconsistent with the help fo `-q` option. If your json is {"dataseturi__uri__contains": "osisaf"} then is serialised directly into dictionary and there is no need to loop over elements:
```suggestion
        cumulative_query = json.loads(arg.query)
```
In case of several queries, the value for input parameter should be:
'{"dataseturi__uri__contains": "osisaf", "source__instrument__short_name__icontains": "AMSR2"}'

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+        max_downloads=int(arg.number_per_day)*(((designated_end-designated_begin).days)+1),
+        use_file_prefix=arg.use_filename_prefix,
+        time_coverage_start__gte=designated_begin,
+        time_coverage_end__lte=designated_end,
+        geographic_location__geometry__intersects=GEOSGeometry(arg.geometry),
+        **cumulative_query
+    )
+    download_manager.download()
+
+def argparse_define():
```
akorosov, line 52: Try to use verb as the first name in the function and don't forget dockstrings.
```suggestion
def parse_args():
""" Parse the script input arguments """ 
```

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    help="The 'wkt' string of geometry which is acceptable by 'GEOSGeometry' of django")
+    parser.add_argument('-c', '--config_file', required=False, type=str,
+    help="The absolute path to the config file that is needed for configuring the downloading "
+    "process. default is the same folder of the 'download.py' file")
+    parser.add_argument('-q', '--query', required=False, type=str,
+    help="""query exposed by user to confine the search result of database for downloading them."
+    "It is a string which must be acceptable by json.loads() to for deserialization of one- or "
+    "multi-criteria limitation."
+    "After deserialization, it must be a list of query that are readable by django filter."
+    "for example a list of elements like {'dataseturi__uri__contains': 'osisaf'}""")
```
akorosov, line 82:
```suggestion
    "for example a list of elements like {"dataseturi__uri__contains": "osisaf"}""")
```
I tried with single quotes and json raised an error and required double quotes.

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
...
+    return parser.parse_args()
+
+if __name__ == "__main__":
+    arg = argparse_define()
```
akorosov, line 86:
```suggestion
    arg = parse_args()
```

setup.py
```diff
@@ -6,14 +6,14 @@
 
 setuptools.setup(
     name="geospaas_processing",
-    version="0.0.4",
-    author="Adrien Perrin",
-    author_email="adrien.perrin@nersc.no",
+    version="1.0.0",
```
akorosov, line 9:
```suggestion
    version="0.1.0",
```

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,87 @@
+"""
+Script for cli of downloading repo. This script accepts the argunemt from sys.args by the
+help of python "argparse" and download the dataset files based on the input arguments.
```
akorosov, line 3:
```suggestion
Download files that are selected from the database using input criteria.
```

Dockerfile_cli
```diff
@@ -0,0 +1,17 @@
+# Dockerfile for building the cli of download with
+# ability of multifolder downloading
+
+
+
```
akorosov, line 5:
```suggestion
# Download CLI
```

Dockerfile_cli
```diff
@@ -0,0 +1,17 @@
...
+ARG BASE_IMAGE=nansencenter/geospaas:2.0.0-slim
+FROM ${BASE_IMAGE} AS base
+RUN pip install --no-cache-dir django-celery-results==1.2 redis graypy
```
akorosov, line 8: Is that needed? setup.py requires only 'django-geo-spaas', 'paramiko', 'scp' which are probably in the geospaas image already.
```suggestion
```

akorosov (changes requested):
Please fix few minor things and add unit tests.

geospaas_processing/cli/download.py
aanersc: I have tested, what you have suggested, the way you create `cumulative_query` with that single line creates a list for `cumulative_query`. We need a dictionary insead of a list to work with the source code of downloading. It is because in the source code of downloading, we use `**` for filtering criteria and `**` is for a dictionary not a list.

geospaas_processing/cli/download.py
aanersc: Done in the next commit

setup.py
aanersc: Done in the next commit

geospaas_processing/cli/download.py
aanersc: Done in the next commit

Dockerfile_cli
aanersc: `django-celery-results` is needed because it is imported in the source code. the last two ones are imported in the test file so they are needed for accomplishing the tests.

Dockerfile_cli
aanersc: Done in the next commit

geospaas_processing/cli/download.py
akorosov: I agree, you need a dict. But this line does create dict:
```python
>>> json.loads('{"dataseturi__uri__contains": "osisaf", "source__instrument__short_name__icontains": "AMSR2"}')
{'dataseturi__uri__contains': 'osisaf', 'source__instrument__short_name__icontains': 'AMSR2'}
```

aanersc: Then I have to change the comment for correct explanation. Sorry for misleading help content. `safe_load` creates a list and json package just serialize and deserialize it.

aanersc: So the help content would be:
```python
"After deserialization, it must be a list of query that are readable by django filter."
    "for example a list of elements like ["dataseturi__uri__contains": "osisaf"]
```

akorosov: OK, so yaml.safe_load returns a list of dicts? Then help string should be `[{"dataseturi__uri__contains": "osisaf"}]`, right? But that makes API more complicated.  Can you simplify API to accept only a dict and then you convert the list to dict in the master script?

By the way. When I run script with `python download.py -q '{a="n"}'` it convert it to `args.query='{a=n}'` (information about quotes is lost). So I have to use backslash in the arguments: `python download.py -q '{a=\"n\"}'` to get correct json: `args.query='{a="n"}'`. Do you have the same? Can you mention that in the help string?

aanersc: Done in the next commit

aanersc:
@akorosov 
@aperrin66 
All addressed in the recent commits as well as introducing the unit tests.

akorosov (dismissed):
Very good!

akorosov:
@aperrin66 , do you want to have a look?

aperrin66:
> @aperrin66 , do you want to have a look?

yes, I will do that

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+    help="The absolute path to the config file that is needed for configuring the downloading "
+    "process. default is the same folder of the 'download.py' file")
+    parser.add_argument('-q', '--query', required=False, type=str,
+    help="""query exposed by user to confine the search result of database for downloading them."
+    "It is a string which must be acceptable by json.loads() to for deserialization of one- or "
+    "multi-criteria limitation."
+    "After deserialization, it must be a list of query that are readable by django filter."
+    "for example a dictionary of elements like "
+    {\"dataseturi__uri__contains\": \"osisaf\", \"source__instrument__short_name__icontains\": \"AMSR2\"}
+    """)
```
aperrin66, line 73: Either use one triple-quoted string or string concatenation, but as it is, here is the printed version of the `help` argument:
```
query exposed by user to confine the search result of database for downloading them."
    "It is a string which must be acceptable by json.loads() to for deserialization of one- or "
    "multi-criteria limitation."
    "After deserialization, it must be a list of query that are readable by django filter."
    "for example a dictionary of elements like "
    {"dataseturi__uri__contains": "osisaf", "source__instrument__short_name__icontains": "AMSR2"}
```

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+        self.assertEqual(arg.begin, '200')
+        self.assertEqual(arg.config_file, '/config_folder/config_file.yml')
+        self.assertEqual(arg.down_dir, '/test_folder/%Y_nh_polstere')
+        self.assertEqual(arg.end, '2020-08-22')
+        self.assertEqual(arg.geometry, 'POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))')
+        self.assertEqual(arg.number_per_day, '100')
+        self.assertEqual(
+            arg.query,
+            '{"dataseturi__uri__contains": "osisaf", "source__instrument__short_name__icontains": "AMSR2"}')
+        # testing the flag performance
```
aperrin66, line 45: This is not performance testing.

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+            '-b', "200",
+            '-e', "2020-08-22",
+            '-r',
+            '-n',"100",
+            '-p',
+            '-g', "POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))",
+            '-c', "/config_folder/config_file.yml",
+            '-q',
+            '{"dataseturi__uri__contains": "osisaf", "source__instrument__short_name__icontains": "AMSR2"}',
+            ]
```
aperrin66, line 31:
```suggestion
        ]
```

geospaas_processing/cli/download.py
aperrin66, line 73: Please indent this properly, this huge block is unreadable.

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+        max_downloads=int(arg.number_per_day)*(((designated_end-designated_begin).days)+1),
+        use_file_prefix=arg.use_filename_prefix,
+        time_coverage_start__gte=designated_begin,
+        time_coverage_end__lte=designated_end,
+        geographic_location__geometry__intersects=GEOSGeometry(arg.geometry),
+        **cumulative_query
+    )
+    download_manager.download()
+
+def parse_args():
```
aperrin66, line 42: Missing docstring.

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+from dateutil.tz import tzutc
+from django.contrib.gis.geos import GEOSGeometry
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'geospaas_processing.settings')
+django.setup()
+from geospaas.catalog.models import Dataset
+import geospaas_processing.downloaders as downloaders
+
+
+def main(arg):
```
aperrin66, line 21: Missing docstring.

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+import sys
+import unittest
+import unittest.mock as mock
+from datetime import datetime
+
+import geospaas_processing.cli.download as cli_download
+import geospaas_processing.downloaders as downloaders
+from dateutil.tz import tzutc
+from django.contrib.gis.geos import GEOSGeometry
+from freezegun import freeze_time
```
aperrin66, line 12: Unused imports and imports order.

```suggestion
import sys
import unittest
import unittest.mock as mock
from datetime import datetime

from dateutil.tz import tzutc
from django.contrib.gis.geos import GEOSGeometry
from freezegun import freeze_time

import geospaas_processing.cli.download as cli_download
```

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+        self.assertTrue(arg.use_filename_prefix)
+        sys.argv.remove('-r')
+        sys.argv.remove('-p')
+        arg = cli_download.parse_args()
+        self.assertFalse(arg.rel_time_flag)
+        self.assertFalse(arg.use_filename_prefix)
+
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.__init__', return_value=None)
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.download')
+    def test_correct_call_json_deserializer(self, mock_downloadmethod, mock_downloadmanagerinit):
```
aperrin66, line 56: please use snake_case

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+        """'json.loads' shall deserializer the whole string that comes after '-q' """
+        arg = cli_download.parse_args()
+        with mock.patch('json.loads') as mock_json:
+            cli_download.main(arg)
+        self.assertIn((
+            '{"dataseturi__uri__contains": "osisaf", "source__instrument__short_name__icontains": "AMSR2"}',),
+            mock_json.call_args)
```
aperrin66, line 63: line too long
```suggestion
        self.assertIn(
            ('{"dataseturi__uri__contains": "osisaf", '
             '"source__instrument__short_name__icontains": "AMSR2"}',),
            mock_json.call_args)
```

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+        sys.argv.remove('-r')
+        sys.argv.remove('-p')
+        arg = cli_download.parse_args()
+        self.assertFalse(arg.rel_time_flag)
+        self.assertFalse(arg.use_filename_prefix)
+
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.__init__', return_value=None)
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.download')
+    def test_correct_call_json_deserializer(self, mock_downloadmethod, mock_downloadmanagerinit):
+        """'json.loads' shall deserializer the whole string that comes after '-q' """
```
aperrin66, line 57:
```suggestion
        """'json.loads' shall deserialize the whole string that comes after '-q' """
```

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+    cumulative_query=json.loads(arg.query) if arg.query else {}
+    if arg.rel_time_flag:
+        designated_begin = datetime.now().replace(tzinfo=tzutc()) + relativedelta(
+            hours=-abs(int(arg.begin)))
+        designated_end = datetime.now().replace(tzinfo=tzutc())
+    else:
+        designated_begin = datetime.strptime(arg.begin, "%Y-%m-%d").replace(tzinfo=tzutc())
+        designated_end = datetime.strptime(arg.end, "%Y-%m-%d").replace(tzinfo=tzutc())
```
aperrin66, line 29: This would be much easier to test if you split this into functions.

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+        with mock.patch('json.loads') as mock_json:
+            cli_download.main(arg)
+        self.assertIsNone(mock_json.call_args)
+
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.__init__', return_value=None)
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.download')
+    def test_correct_call_DownloadManager_without_file_prefix(
+            self, mock_downloadmethod, mock_downloadmanagerinit):
+        """shall return the proper call (with 'use_file_prefix': False) for the case of "
+        "lack of file prefix ('-p') in arguments"""
```
aperrin66, line 81: the name of the method and the docstring don't seem to match.

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+            GEOSGeometry('POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))'),
+            'max_downloads': 30600,
+            'provider_settings_path': '/config_folder/config_file.yml',
+            'time_coverage_end__lte': datetime(2020, 8, 22, 0, 0, tzinfo=tzutc()),
+            'time_coverage_start__gte': datetime(2019, 10, 22, 0, 0, tzinfo=tzutc()),
+            'dataseturi__uri__contains': 'osisaf',
+            'source__instrument__short_name__icontains': 'AMSR2',
+            'use_file_prefix': False
+        },
+            mock_downloadmanagerinit.call_args)
```
aperrin66, line 99:
```suggestion
        }, mock_downloadmanagerinit.call_args)
```

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+            GEOSGeometry('POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))'),
+            'max_downloads': 30600,
+            'provider_settings_path': '/config_folder/config_file.yml',
+            'time_coverage_end__lte': datetime(2020, 8, 22, 0, 0, tzinfo=tzutc()),
+            'time_coverage_start__gte': datetime(2019, 10, 22, 0, 0, tzinfo=tzutc()),
+            'dataseturi__uri__contains': 'osisaf',
+            'source__instrument__short_name__icontains': 'AMSR2',
+            'use_file_prefix': True
+        },
+            mock_downloadmanagerinit.call_args)
```
aperrin66, line 122:
```suggestion
        }, mock_downloadmanagerinit.call_args)
```

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+            GEOSGeometry('POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))'),
+            'max_downloads': 200,
+            'provider_settings_path': '/config_folder/config_file.yml',
+            'time_coverage_end__lte': datetime(2012, 1, 14, 0, 0, tzinfo=tzutc()),
+            'time_coverage_start__gte': datetime(2012, 1, 12, 8, 0, tzinfo=tzutc()),
+            'dataseturi__uri__contains': 'osisaf',
+            'source__instrument__short_name__icontains': 'AMSR2',
+            'use_file_prefix': True
+        },
+            mock_downloadmanagerinit.call_args)
```
aperrin66, line 145:
```suggestion
        }, mock_downloadmanagerinit.call_args)
```

tests/test_cli.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.__init__', return_value=None)
+    @mock.patch('geospaas_processing.downloaders.DownloadManager.download')
+    def test_correct_call_DownloadManager_with_relative_time(
+            self, mock_downloadmethod, mock_downloadmanagerinit):
+        "shall return the proper call for the case of lack of relative time definition in arguments"
+        sys.argv[4] = "40"
+        arg = cli_download.parse_args()
+        with freeze_time("2012-01-14"):
+            cli_download.main(arg)
```
aperrin66, line 132: Isn't installing and using  a new lib a little overkill in this case?
Couldn't you just mock `datetime.now()`?

aperrin66 (changes requested):
A few more things.

Dockerfile_cli
```diff
@@ -0,0 +1,16 @@
+# Download CLI
+
+
+
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

setup.py
```diff
@@ -6,14 +6,14 @@
...
-    author_email="adrien.perrin@nersc.no",
+    version="0.1.0",
+    author=["Adrien Perrin", "Arash Azamifard"],
+    author_email=["adrien.perrin@nersc.no", "arash.azamifard@nersc.no"],
     description="Processing tools for GeoSPaaS",
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://github.com/nansencenter/django-geo-spaas-processing",
-    packages=["geospaas_processing"],
+    packages=["geospaas_processing","geospaas_processing.cli"],
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+import django
+
+from dateutil.relativedelta import relativedelta
+from dateutil.tz import tzutc
+from django.contrib.gis.geos import GEOSGeometry
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'geospaas_processing.settings')
+django.setup()
+from geospaas.catalog.models import Dataset
+import geospaas_processing.downloaders as downloaders
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def main(arg):
+    cumulative_query=json.loads(arg.query) if arg.query else {}
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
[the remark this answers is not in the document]
aanersc: According to my talk with @akorosov , it is enough for this code to understand the dates only(regardless of time of the day). So there is no need to find out the time based on the input. This is implementation is enough and simple.

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+    help="Absolute starting date for download in the format YYYY-MM-DD or (if used together "
+    "with '-r') lag in days relative to today")
+    parser.add_argument('-e', '--end', required=True, type=str,
+    help="Absolute ending date for download in the format YYYY-MM-DD or (if used together "
+    "with '-r') has no influence.")
+    parser.add_argument('-r', '--rel_time_flag', required=False, action='store_true',
+    help="The flag that distinguishes between the two cases of time calcation (1.time-lag from now "
+    "2.Two different points in time) based on its ABSENCE or PRESENCE in the arguments.")
+    parser.add_argument('-n', '--number_per_day', required=False, type=str, default="400",
+    help="limiting number of datasets that are going to be downloaded per day")
```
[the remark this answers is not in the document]
aanersc: So instead of 
`limiting number of datasets that are going to....`
is it better to say
`The upper limit of number of datasets that are going to...`
in this help?

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+        designated_begin = datetime.strptime(arg.begin, "%Y-%m-%d").replace(tzinfo=tzutc())
+        designated_end = datetime.strptime(arg.end, "%Y-%m-%d").replace(tzinfo=tzutc())
+    download_manager = downloaders.DownloadManager(
+        download_directory=arg.down_dir.rstrip(os.path.sep),
+        provider_settings_path=arg.config_file,
+        max_downloads=int(arg.number_per_day)*(((designated_end-designated_begin).days)+1),
+        use_file_prefix=arg.use_filename_prefix,
+        time_coverage_start__gte=designated_begin,
+        time_coverage_end__lte=designated_end,
+        geographic_location__geometry__intersects=GEOSGeometry(arg.geometry),
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+    download_manager.download()
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Process the arguments of entry_point')
+    parser.add_argument('-d', '--down_dir', required=True, type=str,
+    help="Absolute path for downloading files. If path depends on file date, usage of %Y, %m and "
+    "other placeholders interpretable by strftime is accepted")
+    parser.add_argument('-b', '--begin', required=True, type=str,
+    help="Absolute starting date for download in the format YYYY-MM-DD or (if used together "
+    "with '-r') lag in days relative to today")
```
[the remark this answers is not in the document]
aanersc: Yes, it is in hours! Why @akorosov mentioned days in the help text? I don't know. I will fix it to be in hours in the help text.

tests/test_cli.py
aanersc: Yeah, I know that you mean by `performance` in programming.
I will change it with `# testing the flag enumeration`

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+        time_coverage_end__lte=designated_end,
+        geographic_location__geometry__intersects=GEOSGeometry(arg.geometry),
+        **cumulative_query
+    )
+    download_manager.download()
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Process the arguments of entry_point')
+    parser.add_argument('-d', '--down_dir', required=True, type=str,
+    help="Absolute path for downloading files. If path depends on file date, usage of %Y, %m and "
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,78 @@
...
+    "other placeholders interpretable by strftime is accepted")
+    parser.add_argument('-b', '--begin', required=True, type=str,
+    help="Absolute starting date for download in the format YYYY-MM-DD or (if used together "
+    "with '-r') lag in days relative to today")
+    parser.add_argument('-e', '--end', required=True, type=str,
+    help="Absolute ending date for download in the format YYYY-MM-DD or (if used together "
+    "with '-r') has no influence.")
+    parser.add_argument('-r', '--rel_time_flag', required=False, action='store_true',
+    help="The flag that distinguishes between the two cases of time calcation (1.time-lag from now "
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/download.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

geospaas_processing/cli/download.py
aanersc: Done in the next commit

geospaas_processing/cli/download.py
aanersc: Done in the next commit

geospaas_processing/cli/download.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: docstring is corrected. Done in the next commit.

tests/test_cli.py
aanersc: Done in the next commit.

tests/test_cli.py
aanersc: Done in the next commit.

tests/test_cli.py
aanersc: Done in the next commit.

tests/test_cli.py
aanersc: It is not possible to mock `datetime` because `datetime` is a built-in python datatype. We have to use that package in order to have a constant time for testing at different and various times!
It is not possible to mock `datetime.now()`.

geospaas_processing/cli/download.py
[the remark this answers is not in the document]
aperrin66: That's not the problem. It is not a limit per day. It's a safety limit on the number of datasets which can be selected by the given criteria. If the number of datasets selected by the criteria is greater than this limit, an exception is thrown and nothing is downloaded.

tests/test_cli.py
aperrin66: ok

geospaas_processing/cli/download.py
aanersc: Done in the next commit.

aanersc:
@akorosov 
@aperrin66 
As recommended by @aperrin66 , the variation of safety limit based on time is deleted and the safety limit is only passed (conveyed) by this cli.

aperrin66:
@aanersc I didn't notice it before, but the test coverage estimation is broken.

In the .travis.yml file, you need to change the `--source` option of the `coverage run` command from `./geospaas_processing` to `geospaas_processing` (line 51).

aanersc:
@aperrin66 
Done in 6189887

aperrin66 (changes requested):
A few more things, and please check for style issues (lines too long, spaces around operators/arguments, etc...)

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,99 @@
...
+from dateutil.tz import tzutc
+from django.contrib.gis.geos import GEOSGeometry
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'geospaas_processing.settings')
+django.setup()
+import geospaas_processing.downloaders as downloaders
+
+
+def main(arg):
+    "Instantiation and calling the download() method of DownloadManager based on created argparser."
```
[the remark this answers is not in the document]
aanersc: Done

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,99 @@
...
+    cumulative_query = json.loads(arg.query) if arg.query else {}
+    if arg.geometry:
+        cumulative_query.update(
+        {"geographic_location__geometry__intersects": GEOSGeometry(arg.geometry)}
+        )
```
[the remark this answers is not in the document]
aanersc: Done

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,99 @@
...
+        provider_settings_path=arg.config_file,
+        max_downloads=int(arg.number),
+        use_file_prefix=arg.use_filename_prefix,
+        time_coverage_start__gte=designated_begin,
+        time_coverage_end__lte=designated_end,
+        **cumulative_query
+    )
+    download_manager.download()
+
+def find_designated_time(rel_time_flag,begin,end):
```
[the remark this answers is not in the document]
aanersc: Done

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,99 @@
...
+        designated_begin = datetime.now().replace(tzinfo=tzutc()) + relativedelta(
+            hours=-abs(int(begin)))
+        designated_end = datetime.now().replace(tzinfo=tzutc())
+    else:
+        designated_begin = datetime.strptime(begin, "%Y-%m-%d").replace(tzinfo=tzutc())
+        designated_end = datetime.strptime(end, "%Y-%m-%d").replace(tzinfo=tzutc())
+    return designated_begin, designated_end
+
+def parse_args():
+    "creates proper arguments parser with 'argparse' of python."
```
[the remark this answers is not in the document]
aanersc: Done

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,99 @@
...
+    parser.add_argument(
+        '-e', '--end', required=True, type=str,
+        help="Absolute ending date for download in the format YYYY-MM-DD or (if used together "
+        +"with '-r') has no influence.")
+    parser.add_argument(
+        '-r', '--rel_time_flag', required=False, action='store_true',
+        help="The flag that distinguishes between the two cases of time calculation (1.time-lag from "
+        +"now 2.Two different points in time) based on its ABSENCE or PRESENCE in the arguments.")
+    parser.add_argument(
+        '-n', '--number', required=False, type=str, default="400",
```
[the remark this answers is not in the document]
aanersc: Done

geospaas_processing/cli/download.py
```diff
@@ -0,0 +1,99 @@
...
+        +"multi-criteria limitation. "
+        +"After deserialization, it must be a list of query that are readable by django filter."
+        +"for example a dictionary of elements like "
+        +"{\"dataseturi__uri__contains\":\"osisaf\", \"source__instrument__short_name"
+        +"__icontains\":\"AMSR2\"}"
+    )
+    return parser.parse_args()
+
+if __name__ == "__main__":
+    arg = parse_args()
```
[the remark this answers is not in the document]
aanersc: Done

aanersc:
@aperrin66 
Names should be discernible and distinguishable from each other. Therefore, I change this name in order not to get confused with `parser.parse_args` in 5855d9d

aperrin66 (approved):

PR #22, opened 2020-11-06
Issue21 saving the file address after downloading

aanersc:
@aperrin66 
@akorosov 
Since the database is the most important source of information for us and since [as a second reason :)  ] we must take care of staleness of data in the database, I have used `update_or_create` of django for adding the new record in the dataset_uri model. This way, the local address of recent successful download action is stored and we will not be worry about the staleness of the address in the database. And at the same time we have only address which is achievable by using the `get` function of django.

One more test is also added to cover this functionality.

### DIFF
--- a/geospaas_processing/cli/download.py
+++ b/geospaas_processing/cli/download.py
@@ -30,6 +30,7 @@ def main():
         provider_settings_path=arg.config_file,
         max_downloads=int(arg.safety_limit),
         use_file_prefix=arg.use_filename_prefix,
+        store_address=arg.address_storage,
         time_coverage_start__gte=designated_begin,
         time_coverage_end__lte=designated_end,
         **cumulative_query
@@ -79,6 +80,10 @@ def cli_parse_args():
         '-p', '--use_filename_prefix', action='store_true',
         help="The flag that distinguishes between the two cases of having files WITH or WITHOUT "
         + "file prefix when downloaded")
+    parser.add_argument(
+        '-a', '--address_storage', action='store_true',
+        help="The flag that distinguishes between the two cases of whether storing the local "
+        + "address of file in the dataset or not by its its ABSENCE or PRESENCE in the arguments.")
     parser.add_argument(
         '-g', '--geometry', required=False, type=str,
         help="The 'wkt' string of geometry which is acceptable by 'GEOSGeometry' of django")
--- a/geospaas_processing/cli/download.py
+++ b/geospaas_processing/cli/download.py
@@ -30,7 +30,7 @@ def main():
         provider_settings_path=arg.config_file,
         max_downloads=int(arg.safety_limit),
         use_file_prefix=arg.use_filename_prefix,
-        store_address=arg.address_storage,
+        save_path=arg.save_path,
         time_coverage_start__gte=designated_begin,
         time_coverage_end__lte=designated_end,
         **cumulative_query
@@ -81,9 +81,8 @@ def cli_parse_args():
         help="The flag that distinguishes between the two cases of having files WITH or WITHOUT "
         + "file prefix when downloaded")
     parser.add_argument(
-        '-a', '--address_storage', action='store_true',
-        help="The flag that distinguishes between the two cases of whether storing the local "
-        + "address of file in the dataset or not by its ABSENCE or PRESENCE in the arguments.")
+        '-a', '--save_path', action='store_true',
+        help="Save path to local file in the database based on its ABSENCE or PRESENCE.")
     parser.add_argument(
         '-g', '--geometry', required=False, type=str,
         help="The 'wkt' string of geometry which is acceptable by 'GEOSGeometry' of django")
--- a/geospaas_processing/cli/download.py
+++ b/geospaas_processing/cli/download.py
@@ -83,7 +83,7 @@ def cli_parse_args():
     parser.add_argument(
         '-a', '--address_storage', action='store_true',
         help="The flag that distinguishes between the two cases of whether storing the local "
-        + "address of file in the dataset or not by its its ABSENCE or PRESENCE in the arguments.")
+        + "address of file in the dataset or not by its ABSENCE or PRESENCE in the arguments.")
     parser.add_argument(
         '-g', '--geometry', required=False, type=str,
         help="The 'wkt' string of geometry which is acceptable by 'GEOSGeometry' of django")
--- a/geospaas_processing/downloaders.py
+++ b/geospaas_processing/downloaders.py
@@ -275,6 +275,8 @@ class DownloadManager():
         Returns the downloaded file path if the download succeeds, an empty string otherwise.
         """
         for dataset_uri in dataset.dataseturi_set.all():
+            if dataset_uri.service==geospaas.catalog.managers.LOCAL_FILE_SERVICE:
+                continue
             # Get the extra settings for the provider
             dataset_uri_prefix = "://".join(requests.utils.urlparse(dataset_uri.uri)[0:2])
             # Find provider settings
--- a/geospaas_processing/downloaders.py
+++ b/geospaas_processing/downloaders.py
@@ -275,7 +275,7 @@ class DownloadManager():
         Returns the downloaded file path if the download succeeds, an empty string otherwise.
         """
         for dataset_uri in dataset.dataseturi_set.all():
-            if dataset_uri.service==geospaas.catalog.managers.LOCAL_FILE_SERVICE:
+            if dataset_uri.service == geospaas.catalog.managers.LOCAL_FILE_SERVICE:
                 continue
             # Get the extra settings for the provider
             dataset_uri_prefix = "://".join(requests.utils.urlparse(dataset_uri.uri)[0:2])
@@ -323,7 +323,7 @@ class DownloadManager():
                         if self.save_path:
                             dataset.dataseturi_set.get_or_create(
                                 dataset=dataset,
-                                uri = os.path.join(download_directory, file_name),
+                                uri = os.path.join(os.path.realpath(download_directory), file_name),
                                 )
                         LOGGER.info("Successfully downloaded dataset %d to %s",
                                     dataset.pk, file_name)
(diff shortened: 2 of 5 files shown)


geospaas_processing/downloaders.py
```diff
@@ -317,6 +318,11 @@ def download_dataset(self, dataset, download_directory):
                         f"Could not write the dowloaded file to {error.filename}") from error
                 else:
                     if downloaded:
+                        if self.store_address:
+                            dataset.dataseturi_set.update_or_create(
+                                service='DOWNLOAD_DIR', dataset=dataset,
+                                defaults={"uri": os.path.join(download_directory, file_name)}
+                                )
```
akorosov, line 325:
```suggestion
                        if self.save_path:
                            dataset.dataseturi_set.update_or_create(
                                dataset=dataset,
                                uri = os.path.join(download_directory, file_name),
                                )
```

aanersc: @aperrin66 
@akorosov 
If we use `get_or_create` then we may have several addresses stored in the database that happens for the cases of changing the downloading directory when we run the downloading cli. 

What is the most valid address for copying the file? obviously the last successful attempt of download. So copying must be down with the source of last downloaded ones. Old addresses might be removed!

Otherwise, if we use `update_or_create` there is always one record in the dataset_uri model which points to the most recent location of the file. This way we can use `get` of django and not concern about the duplicating addresses in the database. 

The way I wrote the `update_or_create` is the way that it searches based on the service='DOWNLOAD_DIR' and then it updates the address if necessary. So, "uri" must come as a default for `update_or_create` in order to be updated if necessary: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#update-or-create

This suggestion does not doing that algorithm.

aperrin66: Maybe this is not the place to check for the validity of existing DatasetURIs. We talked about having a separate checking mechanism, maybe a script executed regularly which would remove the stale DatasetURIs.
Considering this, maybe you can just add a new DatasetURI here and not bother with the existing ones.

geospaas_processing/cli/download.py
```diff
@@ -30,6 +30,7 @@ def main():
         provider_settings_path=arg.config_file,
         max_downloads=int(arg.safety_limit),
         use_file_prefix=arg.use_filename_prefix,
+        store_address=arg.address_storage,
```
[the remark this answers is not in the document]
aanersc: DOne in the next commit.

geospaas_processing/cli/download.py
```diff
@@ -79,6 +80,10 @@ def cli_parse_args():
         '-p', '--use_filename_prefix', action='store_true',
         help="The flag that distinguishes between the two cases of having files WITH or WITHOUT "
         + "file prefix when downloaded")
+    parser.add_argument(
+        '-a', '--address_storage', action='store_true',
+        help="The flag that distinguishes between the two cases of whether storing the local "
+        + "address of file in the dataset or not by its ABSENCE or PRESENCE in the arguments.")
```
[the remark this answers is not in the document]
aanersc: DOne in the next commit.

geospaas_processing/downloaders.py
```diff
@@ -238,7 +238,7 @@ class DownloadManager():
     }
 
     def __init__(self, download_directory='.', provider_settings_path=None, max_downloads=100,
-                 use_file_prefix=True, **criteria):
+                 use_file_prefix=True, store_address=False, **criteria):
```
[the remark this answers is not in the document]
aanersc: DOne in the next commit.

geospaas_processing/downloaders.py
```diff
@@ -251,6 +251,7 @@ def __init__(self, download_directory='.', provider_settings_path=None, max_down
             raise DownloadError("No dataset matches the search criteria")
         self.download_folder = download_directory
         self.use_file_prefix = use_file_prefix
+        self.store_address = store_address
```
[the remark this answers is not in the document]
aanersc: DOne in the next commit.

geospaas_processing/downloaders.py
aanersc: DOne in the next commit.

aanersc:
@aperrin66 @akorosov 
If there are a local address stored as dataset_uri then, there should not be any attempt done in the downloading based on that! We have no downloader for a local file address!   :)
This needs a tiny coding!

And Also for tests need one more test:
So I create a second fake uri and a local address for one of the datasets in the fixtures and test it the same way of testing the dataset in the fixture having the fake_uri. The the way the test is written it the same but it still must create the download error regardless of existance or absence of the local address. The reason is that the download error is coming from "a corrupted url"("a fake url").

geospaas_processing/downloaders.py
```diff
@@ -274,6 +275,8 @@ def download_dataset(self, dataset, download_directory):
         Returns the downloaded file path if the download succeeds, an empty string otherwise.
         """
         for dataset_uri in dataset.dataseturi_set.all():
+            if dataset_uri.service==geospaas.catalog.managers.LOCAL_FILE_SERVICE:
```
aperrin66, line 278: Spaces around `==`

tests/test_cli.py
```diff
@@ -65,8 +69,7 @@ def test_correct_call_json_deserializer(self, mock_download_method, mock_downloa
 
     @mock.patch('geospaas_processing.downloaders.DownloadManager.__init__', return_value=None)
     @mock.patch('geospaas_processing.downloaders.DownloadManager.download')
-    def test_lack_of_calling_json_deserializer_when_no_query_appears(
-            self, mock_download_method, mock_download_manager_init):
+    def test_lack_of_calling_json_deserializer_when_no_query_appears(self, mock_download_method, mock_download_manager_init):
```
aperrin66, line 72: This line is now too long.

geospaas_processing/downloaders.py
```diff
@@ -317,6 +320,11 @@ def download_dataset(self, dataset, download_directory):
                         f"Could not write the dowloaded file to {error.filename}") from error
                 else:
                     if downloaded:
+                        if self.save_path:
+                            dataset.dataseturi_set.get_or_create(
+                                dataset=dataset,
+                                uri = os.path.join(download_directory, file_name),
```
aperrin66, line 326: You should store an absolute path. `download_directory` is not necessarily one.
You can use https://docs.python.org/3/library/os.path.html#os.path.realpath

tests/test_downloaders.py
```diff
@@ -384,6 +385,19 @@ def test_the_correct_call_of_check_and_download_url_without_file_prefix(self):
...
+        Test that address of downloaded file is added to the dataseturi model.
+        """
+        download_manager = downloaders.DownloadManager(save_path=True)
+        dataset = Dataset.objects.get(pk=3)
+        with mock.patch.object(downloaders.HTTPDownloader, 'check_and_download_url') as mock_dl_url:
+            mock_dl_url.return_value = ('test.nc', True)
+            download_manager.download_dataset(dataset, 'testing_value')
+            self.assertEqual(dataset.dataseturi_set.filter(
+                                dataset=dataset,service=LOCAL_FILE_SERVICE)[0].uri,
+                            'testing_value/test.nc')
```
aperrin66, line 399: As stated in another comment, the stored path should be absolute.

tests/test_downloaders.py
```diff
@@ -449,7 +463,17 @@ def check_and_download_url_side_effect(url, download_dir, file_prefix='', **kwar
...
+        """Test that `download_dataset` raises a DownloadError exception if the download failed"""
+        download_manager = downloaders.DownloadManager()
+        dataset = Dataset.objects.get(pk=2)
+        with mock.patch.object(downloaders.HTTPDownloader, 'check_and_download_url') as mock_dl_url:
+            mock_dl_url.side_effect = downloaders.DownloadError
+            with self.assertRaises(downloaders.DownloadError):
+                with self.assertLogs(downloaders.LOGGER, logging.WARNING):
+                    download_manager.download_dataset(dataset, '')
+
+    def test_download_dataset_without_local_link_fails_because_of_corrupted_uri(self):
```
aperrin66, line 476: I don't get why you added that test. The point of this test was to check that errors happening during the call to `check_and_download_url()` are correctly propagated. It has nothing to do with local URIs.

If you want to check that we don't attempt to download local URIs, you should write a test dedicated to that.

aperrin66, line 476: The names you chose are confusing. Problems could happen for a number of reasons: an expired certificate, a wrong URI, the remote site is down for maintenance, etc... I think you should remove the `because_of_corrupted_uri` part.

aperrin66 (changes requested):

geospaas_processing/downloaders.py
aanersc: Done in the next commit

tests/test_downloaders.py
aanersc: I have replaced `corrupted` with `unresponsive`.

aanersc: These two test are actually real cases that simulates the cases in reality. They are kind of integration tests that can we nominate them. It does matter to have all kind of things in the dataset_uri model at the same time and the errors are properly raised!

aperrin66: That is still too specific. As I wrote before, there can be multiple reasons for these errors.

aanersc: Done in the next commit. Another test is added to the suite.

tests/test_downloaders.py
aanersc: Done in the next commit

geospaas_processing/downloaders.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

geospaas_processing/downloaders.py
aanersc: Done in the next commit

tests/test_downloaders.py
aanersc: I remove the last part of the name

aanersc:
@aperrin66 
All addressed.

akorosov (approved):
Looks good to me!

aperrin66 (approved):

PR #24, opened 2020-11-10
Issue23 copy cli

aanersc:
@aperrin66 
@akorosov 
copy cli and some prelimiary tests for it are added here in this PR. The test development is still needed. However, the CLI is in its final version and you are kindly asked to review it.


aanersc:
@aperrin66
@akorosov

Now the tests are complete for copy cli!!!

tests/test_cli.py
```diff
@@ -182,3 +191,169 @@ def test_find_designated_time_function(self):
...
+        sys.argv = [
+            "",
+            '-b', "2018-04-01",
+            '-e', "2018-04-09",
+            '-f', '-l'
+        ]
+        temp_directory_dst = tempfile.TemporaryDirectory()
+        sys.argv.append('-d')
+        sys.argv.append(temp_directory_dst.name)
+        os.mkdir("/testing_folder")
```
aperrin66, line 335: Don't create files and directories under `/`.
You created a temporary directory, use it: you can have both your source and destination folders in it.

tests/test_cli.py
```diff
@@ -182,3 +191,169 @@ def test_find_designated_time_function(self):
...
+        Path("/testing_folder/testing_file.test").touch()
+        cli_copy.main()
+        self.assertTrue(os.path.islink(temp_directory_dst.name+'/testing_file.test'))
+        cli_copy.main()  # assertion of completion of copy for the second time without error
+        self.assertTrue(os.path.islink(temp_directory_dst.name+'/testing_file.test'))
+        shutil.rmtree("/testing_folder")
+
+    def test_copying_accomplishment_and_consistency(self):
```
aperrin66, line 343: This test method has the same name as the previous one, so the first one is never executed.

aperrin66 (changes requested):

geospaas_processing/cli/README.md
```diff
@@ -0,0 +1,17 @@
+# Overal table for showing the usage of short-form of arguments of all CLIs individually
```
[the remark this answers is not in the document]
aanersc: Yeah, you are right. I will put it after `Dependencies` section because people may need to see it many times when they use CLIs.

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Copy the files based on the addressed stored in database to the destination folder.
+    The destination folder as well as criteria for finding the desired files are input of this
+    function.
+    """
+    arg = cli_parse_args()
+    cumulative_query = json.loads(arg.query) if arg.query else {}
+    if arg.geometry:
+        cumulative_query['geographic_location__geometry__intersects'] = GEOSGeometry(arg.geometry)
+    designated_begin, designated_end = download_cli.find_designated_time(
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
+"""
+Download files that are selected from the database using input criteria.
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+    designated_begin, designated_end = download_cli.find_designated_time(
+        arg.rel_time_flag, arg.begin, arg.end)
+    cumulative_query['time_coverage_start__gte'] = designated_begin
+    cumulative_query['time_coverage_end__lte'] = designated_end
+    datasets = Dataset.objects.filter(**cumulative_query)
+    for dataset in datasets:
+        try:
+            source_paths = dataset.dataseturi_set.filter(service=LOCAL_FILE_SERVICE)
+        except IndexError:
+            continue
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+        for source_path in source_paths:
+            if path.isfile(source_path.uri):
+                if arg.link:
+                    try:
+                        os.symlink(dst=arg.destination_path.rstrip(os.path.sep)
+                                   + os.path.sep+ntpath.basename(source_path.uri),
+                                   src=source_path.uri)
+                    except FileExistsError:
+                        pass
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+        cumulative_query['geographic_location__geometry__intersects'] = GEOSGeometry(arg.geometry)
+    designated_begin, designated_end = download_cli.find_designated_time(
+        arg.rel_time_flag, arg.begin, arg.end)
+    cumulative_query['time_coverage_start__gte'] = designated_begin
+    cumulative_query['time_coverage_end__lte'] = designated_end
+    datasets = Dataset.objects.filter(**cumulative_query)
+    for dataset in datasets:
+        try:
+            source_paths = dataset.dataseturi_set.filter(service=LOCAL_FILE_SERVICE)
+        except IndexError:
```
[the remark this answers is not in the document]
aanersc: I have replaced it with `exists()`

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+            continue
+        for source_path in source_paths:
+            if path.isfile(source_path.uri):
+                if arg.link:
+                    try:
+                        os.symlink(dst=arg.destination_path.rstrip(os.path.sep)
+                                   + os.path.sep+ntpath.basename(source_path.uri),
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+                                   src=source_path.uri)
+                    except FileExistsError:
+                        pass
+                else:
+                    shutil.copy(src=source_path.uri, dst=arg.destination_path)
+                if arg.flag_file:
+                    with open(arg.destination_path.rstrip(os.path.sep)
+                              + os.path.sep + ntpath.basename(source_path.uri)
+                              + ".flag", "w+") as flag_file:
```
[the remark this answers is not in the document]
aanersc: Done

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: I use ntpath because some users (in future) may come from the windows world and always use windows! By using ntpath are unconcerned that the address separation character does not cause a problem for those kind of users.

tests/test_cli.py
```diff
@@ -1,17 +1,25 @@
...
+import django.test
+import geospaas_processing.cli.copy as cli_copy
+import geospaas_processing.cli.download as cli_download
 from dateutil.tz import tzutc
 from django.contrib.gis.geos import GEOSGeometry
 from freezegun import freeze_time
-
-import geospaas_processing.cli.download as cli_download
+from geospaas.catalog.models import Dataset
+from geospaas.catalog.managers import LOCAL_FILE_SERVICE
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

tests/test_cli.py
```diff
@@ -182,3 +191,169 @@ def test_find_designated_time_function(self):
...
+        arg = cli_copy.cli_parse_args()
+        self.assertEqual(arg.begin, '200')
+        self.assertEqual(arg.destination_path, '/test_folder/')
+        self.assertEqual(arg.end, '2018-11-18')
+        self.assertEqual(arg.geometry, 'POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))')
+        self.assertEqual(arg.type, 'test_type')
+        self.assertEqual(arg.query,
+                         '{"dataseturi__uri__contains": "osisaf", '
+                         + '"source__instrument__short_name__icontains": "AMSR2"}')
+        # testing the flag enumeration
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

tests/test_cli.py
```diff
@@ -182,3 +191,169 @@ def test_find_designated_time_function(self):
...
+        self.assertTrue(arg.rel_time_flag)
+        self.assertTrue(arg.flag_file)
+        self.assertTrue(arg.link)
+        sys.argv.remove('-r')
+        sys.argv.remove('-f')
+        sys.argv.remove('-l')
+        arg = cli_copy.cli_parse_args()
+        self.assertFalse(arg.rel_time_flag)
+        self.assertFalse(arg.flag_file)
+        self.assertFalse(arg.link)
```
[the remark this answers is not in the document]
aanersc: Because first we have to assert the flag is conveyed corrected to this step of workflow. And then by removing the flag once again make ourselves secured of falseness of this values. So two aspects must be tested apart from each other.

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
+"""
+Download files that are selected from the database using input criteria.
+"""
+import argparse
+import json
+import ntpath
+import os
+from os import path
```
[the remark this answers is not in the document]
aanersc: No, I have intentionally used it this way in order not to get confused in the tests for the definition of `path.is_file` which is used in the assertion of the tests. This way is much more clear with the correct mocking of usage (instead of definition) of path.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: I replace that with os.join

OS is never patched! it is the `path` which is patched!

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+        arg.rel_time_flag, arg.begin, arg.end)
+    cumulative_query['time_coverage_start__gte'] = designated_begin
+    cumulative_query['time_coverage_end__lte'] = designated_end
+    datasets = Dataset.objects.filter(**cumulative_query)
+    for dataset in datasets:
+        try:
+            source_paths = dataset.dataseturi_set.filter(service=LOCAL_FILE_SERVICE)
+        except IndexError:
+            continue
+        for source_path in source_paths:
```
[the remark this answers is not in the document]
aanersc: No. This way, we are sure that even if only one local address is valid(I mean is placed there and ready to copy), then copy action will be accomplished. If there are some addresses in the database but the file is moved or maybe dabase is being manipulated by other workflows, this way of coding makes us assure that the file is copied.

Otherwise, as soon as a invalid local address is found in the database, the file is not copied.

Fortunately, the `shutil.copy` does not create duplicate of files if we run the code several times. This means we have idempotency.

tests/test_cli.py
```diff
@@ -182,3 +191,169 @@ def test_find_designated_time_function(self):
...
+        ]
+        temp_directory = tempfile.TemporaryDirectory()
+        sys.argv.append('-d')
+        sys.argv.append(temp_directory.name)
+        cli_copy.main()
+        fd = os.open(temp_directory.name + '/testing_file.test.flag', os.O_RDONLY)
+        self.assertEqual(os.read(fd, 200),
+                         b"type: test_type\nurl: https://scihub.copernicus.eu/apihub/odata/v1/Produ"
+                         + b"cts('6127111d-c9bd-4689-bab5-412dd39e1e81')/$value\nurl: https://scihu"
+                         + b"b.copernicus.eu/the_second_fakeurl\n")
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aperrin66: That's not a valid justification. It's not a problem to mock `os.path`.
I won't argue about that again, please read https://docs.python.org/3/library/unittest.mock.html#where-to-patch if you have doubts.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aperrin66: My comment has nothing to do with patching.
If you use `import os.path`, you should use `os.path` in the rest of your code (I am not talking about the tests).
If you use `from os import path`, you should use `path` in the rest of your code.

tests/test_cli.py
```diff
@@ -182,3 +191,169 @@ def test_find_designated_time_function(self):
...
+            mock_symlink.call_args_list)
+        for dataset in Dataset.objects.all():
+            dataset.dataseturi_set.filter(
+                uri='/new_loc_add', dataset=dataset, service=LOCAL_FILE_SERVICE).delete()
+
+    @mock.patch('os.symlink')
+    @mock.patch('geospaas_processing.cli.copy.path')
+    def test_correct_content_of_flag_file(self, mock_isfile, mock_link):
+        """ flag file should contain this 'type: test_type' information """
+        mock_isfile.isfile.return_value = True
```
aperrin66, line 306: What you are doing here is:
- you patch `geospaas_processing.cli.copy.path`, creating a Mock object
- you then edit the `isfile` property of this Mock object

This means that you are patching the whole `os.path` module, instead of just the `isfile` function, which is what you need.

So please patch `geospaas_processing.cli.copy.path.isfile` directly.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aperrin66: Using `os.path` will automatically choose the right implementation of the module (`posixpath` or `ntpath`).
You should not use `posixpath` or `ntpath` directly except in very specific cases. This is not one of them.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aperrin66: `shutil.copy()` replaces the destination file if it already exists, so it would be best to stop the loop after the first successful copy.

tests/test_cli.py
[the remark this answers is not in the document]
aperrin66: Then why don't you test the options which take an argument in the same way?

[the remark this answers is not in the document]
aanersc: What do you mean? these are flags are change the behaviour by their presence and absence. The others are tested by their values! Is there any more tests needed?

tests/test_cli.py
```diff
@@ -182,3 +191,169 @@ def test_find_designated_time_function(self):
...
+            '-e', "2018-11-18",
+            '-r',
+            '-f',
+            '-l',
+            '-g', "POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))",
+            '-t', 'test_type'
+        ]
+        with mock.patch('json.loads') as mock_json:
+            cli_copy.main()
+        self.assertIsNone(mock_json.call_args)
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: I don't know that! thanks. Done in the next commit.

tests/test_cli.py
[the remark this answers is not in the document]
aperrin66: What I mean is: what happens if the others are absent?

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: As you have mentioned, I used `import os.path`, and I use `os.path` in the rest of my code. But exactly in one line:
https://github.com/nansencenter/django-geo-spaas-processing/blob/63a5948d7a6b2e0e6f931cb4756fb489ebbb3426/geospaas_processing/cli/copy.py#L42

I have used `path.is_file` and just patch this one in the tests. The reason is that this line is the only line which is needed to be patched during the tests. Everything is clear with the tests now! If we want to code that approach(which is true). The tests become confusing!

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
akorosov: I don't see this is done. `download_cli` is still imported from `geospaas_processing.cli.download`.

But it should be imported from a separate `utils` module as suggested by adrien.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
akorosov: I agree. Better to add a `break`

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
akorosov: Where? Did you forget to push?

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
akorosov: You can patch "os.path.isfile" and then the rest of the `os` or `os.path` is not affected.

tests/test_cli.py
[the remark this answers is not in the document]
aanersc: Others are not working by their absence or presence. They are working by their values, which are tested in the above lines! These are the one that are resulting in different situation by their absence and presence and both `absence` and `presence` situation of them must be approved to be false or true.

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+                if arg.flag_file:
+                    with open(arg.destination_path.rstrip(os.path.sep)
+                              + os.path.sep + ntpath.basename(source_path.uri)
+                              + ".flag", "w+") as flag_file:
+                        flag_file.write(f"type: {arg.type}" + os.linesep)
+                        for urlname in dataset.dataseturi_set.exclude(service=LOCAL_FILE_SERVICE):
+                            flag_file.write(f"url: {urlname.uri}" + os.linesep)
+
+
+def cli_parse_args():
```
akorosov, line 61: This function largerly repeats cli_parse_args from download.py
What's worse, in some places it is somewhat similar. It either should be exactly the same (and then code should be moved to cli_utils.py) or it should be distinctly different. For example,
`-d` is present in both. But in one it turnes into `destination_path` and in another into `down_dir`. this is very confusing. The best would be to call both parameters `destination` and use `-d` and move common code of parser initialisation to cli_utils.py.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: Done, I have added one more line based on your suggestion to prevent `shutil.copy` from replacing!

akorosov (changes requested):
Looks like some files are not pushed yet. It is said in some comments that suggestions from Adrien are addressed but it is not visible in the code.
Please:
1. Implement all Adriens suggestions (also regarding os.path)
1. Address my questions
1. Push the latest changes

tests/test_cli.py
aanersc: The fixture are installed in the database. The address in the fixture is constant and we can change it based various execution of testing suite. So it is necessary to create a real file in a real folder for testing purposes!

We can change the destination folder based on the tempfolder at the time of execution of the tests. But this is not doable for the source address because source address comes from the database.

tests/test_cli.py
aanersc: Thanks for scrutinising comment! I have tested them with other names! I forget to change the name!

tests/test_cli.py
aperrin66: Then at least create it in `/tmp`.

aanersc:
@akorosov 
https://github.com/nansencenter/django-geo-spaas-processing/pull/24#discussion_r527679866

I have not finished with all comments of adrien, if I do I push them and you will see the new version of code. When I wrote `Done in the next commit`, it means There some other comments are taking into account.

aanersc:
@akorosov 
If we go for patching the definition, the tests become ugly! This is a serious warning that I have notice for your insistence of pathing the 'os.isfile' in the tests.

tests/test_cli.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: It does not work with patching `geospaas_processing.cli.copy.path.isfile` in the decorator. We need that Mock object for the testing execution working properly. To anyway, as we have discussed internally, I go with your approach of patching.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: I have done it in your way. Done in the next commit.

tests/test_cli.py
aanersc: Done in the next commit.

geospaas_processing/cli/copy.py
aanersc: Done in the next commit.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: Discussed. Done in the next commit.

geospaas_processing/cli/copy.py
```diff
@@ -0,0 +1,108 @@
...
+                        pass
+                else:
+                    shutil.copy(src=source_path.uri, dst=arg.destination_path)
+                if arg.flag_file:
+                    with open(arg.destination_path.rstrip(os.path.sep)
+                              + os.path.sep + ntpath.basename(source_path.uri)
+                              + ".flag", "w+") as flag_file:
+                        flag_file.write(f"type: {arg.type}" + os.linesep)
+                        for urlname in dataset.dataseturi_set.exclude(service=LOCAL_FILE_SERVICE):
+                            flag_file.write(f"url: {urlname.uri}" + os.linesep)
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

aanersc:
@aperrin66 @akorosov 
Since there are big changes in the code especially because of this comment https://github.com/nansencenter/django-geo-spaas-processing/pull/24#discussion_r526029824 of @aperrin66 , I make a single commit for it! Now the common parts in the util.py.

I do my best to make it ready a further development of redis if needed. I try to make it generic as much as possible to be executable both from CLI and from the redis in the future.

The algorithm is the same, I just rewrite it in a more modularized manner. 
@aperrin66 @akorosov 
All of our discussion is considered in this version. Now it is ready for review.

geospaas_processing/copiers.py
```diff
@@ -0,0 +1,72 @@
...
+            flag_file.write(f"type: {type_in_flag_file}" + os.linesep)
+            for urlname in dataset.dataseturi_set.exclude(service=LOCAL_FILE_SERVICE):
+                flag_file.write(f"url: {urlname.uri}" + os.linesep)
+
+    def file_or_symlink_copy(self, source_paths, dataset):
+        """copy the file or a symlink of the file of dataset based on its stored local address in
+        the database."""
+        for source_path in source_paths:
+            if os.path.isfile(source_path.uri):
+                sopposed_full_address = os.path.join(self.destination_path.rstrip(os.path.sep),
```
akorosov, line 42:
```suggestion
                destination_filename = os.path.join(self.destination_path,
```
1. please rename `sopposed_full_address` everywhere
2. `os.path.join('a', 'b')` and `os.path.join('a/', 'b')` produce the same result. That what `join` is for.

README.md
```diff
@@ -10,6 +10,24 @@ It is composed of:
 
 The processing modules can either be run as standalone code or asynchronously as Celery tasks.
 
+---
+# Overal table for showing the usage of short-form of arguments of all CLIs individually
+| Argument short form | Download CLI            | Copy CLI
+| ------------------- | --------------------    | --------------
+| '-d'                | '--down_dir'            | '--destination_path'
```
akorosov, line 17:
```suggestion
| '-d'                | '--destination_dir'            | '--destination_dir'
```
Also move adding the `destination_dir` argument to `util.parse_common_args`

geospaas_processing/copiers.py
```diff
@@ -0,0 +1,72 @@
...
+from geospaas.catalog.models import Dataset
+from geospaas.catalog.managers import LOCAL_FILE_SERVICE
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'geospaas_processing.settings')
+django.setup()
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Copier():
```
akorosov, line 16:
```suggestion
class CopyManager():
```
Maybe? To be consistent with DownloadManagers

tests/test_cli.py
```diff
@@ -176,10 +185,176 @@ def test_correct_call_DownloadManager_with_relative_time(
...
+        """ the copied file(s) shall be copied at the destination folder """
+        sys.argv = [
+            "",
+            '-b', "2018-04-01",
+            '-e', "2018-04-09",
+            '-d', "/dst_folder/"
+        ]
+        for dataset in Dataset.objects.all():
+            dataset.dataseturi_set.get_or_create(
+                uri='/new_loc_add', dataset=dataset, service=LOCAL_FILE_SERVICE)
```
akorosov, line 270: That should be in fixtures.

tests/test_cli.py
```diff
@@ -176,10 +185,176 @@ def test_correct_call_DownloadManager_with_relative_time(
...
+            '-f',
+            '-l',
+            '-g', "POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))",
+            '-t', 'test_type'
+        ]
+        with mock.patch('json.loads') as mock_json:
+            cli_copy.main()
+        mock_json.assert_not_called()
+
+    @mock.patch('os.path.isfile', side_effect=[True, False, True, False])
```
akorosov, line 256: Nice!

tests/test_cli.py
```diff
@@ -176,10 +185,176 @@ def test_correct_call_DownloadManager_with_relative_time(
...
+        sys.argv = [
+            "",
+            '-b', "2018-04-01",
+            '-e', "2018-04-09",
+            '-f', '-l'
+        ]
+        temp_directory_dst = tempfile.TemporaryDirectory()
+        sys.argv.append('-d')
+        sys.argv.append(temp_directory_dst.name)
+        Path("/tmp/testing_file.test").touch()
```
akorosov, line 338: Shouldn't `"/tmp/testing_file.test"` be `tempfile.TemporaryFile`?

akorosov (changes requested):
Looks nice!
I could suggest only few renamings and minor improvements in the tests.

README.md
aanersc: `down_dir` concept is different from `destination_path` concept, so they can not be merged into one common thing! If we want to merge, then what should I right for in the help of the argument so that the users of both CLIs are not confused? 

We also need to change some other files (templates, configurations) if we want to change the name! I think it is better not to merge them because it becomes obscure and may become confusing.

geospaas_processing/copiers.py
aanersc: copying action is one type of action, unlike the download action which uses an specific kind of downloader and manages by download manager. It is better to keep this name in hormony with downloader.py. Respectfully, what is your opinion @aperrin66 ? Am I understand the downloader.py correctly? Am I concluding the correct way from my understanding?

tests/test_cli.py
aanersc: Discussed beforehand! Here is the start of discussion: https://github.com/nansencenter/django-geo-spaas-processing/pull/24#discussion_r526108351

aanersc: Keep it simple! We have something else in our mind when we wrote this test! No need to engage ourselves with TemporaryFile.

On the other hand, TemporaryFile will be destroyed as soon as it is closed as its doc revealed https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile!  which is not our case here!

tests/test_cli.py
aanersc: Fixtures are designed carefully for having all possible situations in them, nothing more and nothing less. Here we have 3 datasets in the fixture which have below list of dataset_uri (numbers of datasets maybe not correct in the list below):
3. healthy url
1.healthy url **_plus_** fake url
2.healthy url **_plus_** the second fake url **_plus_** local address

In these three comment line I add another local address named `/new_loc_add` for **all** dataset! This means if we move it to fixtures, the fixtures will be polluted and we have to change a lot of tests accordingly.

The reason why I add that these lines is that maybe someone download ALL the datasets once again in another folder based on his/her choice, so in that case, copy should be able to access and search both of addresses stored in the database. 

So it is better to have this code here because they are only belong to this test! only!

I will change the docstring of the test based on this comment of yours.

geospaas_processing/copiers.py
aanersc: Please have a look:
image

The third and the fourth case are the cases that I am worried about! It is better to have these rstrip and also lstrip in the code in order not to deal with this simple issue later on!

aanersc: The name has been changed. Done in the next commit.

geospaas_processing/copiers.py
aperrin66: I'm not sure either of your solutions is well suited for this case.
We use Dowloaders and a DownloadManager because we need different implementations of the downloading action depending on the type of dataset. This is not the case here, because copying is always done the same way (or maybe you could have one copier for "real" copies, and one for symlinks, but that seems a little over the top).

To be honest, I'm not even sure if object oriented programming is necessary here.

aperrin66: I just had a thought: couldn't the copying mechanism be implemented as a Downloader? It's just a particular case were the files are located on the same server. This would remove the need for a separate CLI, config file, etc...

aanersc:
@akorosov 
All of your discussions are addressed in 645befc.

tests/test_cli.py
aperrin66: You can use `tempfile.TemporaryFile` as a context manager (`with temfile.TemporaryFile...`) to keep it open as long as you need.

geospaas_processing/copiers.py
aanersc: Defining a new workflow for differentiating the symlink from file itself is not necessary because there is only one line of code which varies between them! only one line! So, there is no need to do that.

Yeah, you are right about the OOP. I have done this task with some functions that are only gathered together in a class without any constructer! Then I saw that the arguments of the functions are a lot and they are repeat several times, so I decided to you use OOP for `simplification of implementation`.

By the way, OOP helps us to make sure that all of these:
```python
        self.type_in_flag_file = type_in_flag_file
        self.destination_path = destination_path
        self.flag_file_request = flag_file_request
        self.link_request = link_request
        self.datasets = Dataset.objects.filter(**criteria)
```
Are not changing during one copy action! This makes it robust and less error prone for the workflow! It is because they are constant during the execution of functions!



Regarding https://github.com/nansencenter/django-geo-spaas-processing/pull/24#discussion_r529387018:
This is one of those thunderbolt ideas that needs time and effort! I think it worth it. BUT please let us finish this PR and make a version number after it. And then we will implement this idea! OK?

I also appreciated if you two tell to how to name that class!!!  :)

geospaas_processing/copiers.py
aperrin66: I agree with Anton on the second point.

`a//b` is equivalent to `a/b`.
`/b` is an absolute path, which should only be used as first argument to `os.path.join()`. This situation will never happen because you use `os.path.basename()` on the second argument.

tests/test_cli.py
aanersc: When we use a temporary folder which contains that file, then why should we bother ourselves two code with a context manager? the subject is not that file! the subject is something else! Now the test is written in a `without indentation` manner and it is simpler to understand than using a context manager.

geospaas_processing/copiers.py
aanersc: Done in the next commit.

geospaas_processing/copiers.py
aanersc: Maybe it is better for me to write that attributes in the constructor with capitalised letters based on the pep8.

aanersc:
@aperrin66 
@akorosov 
All the discussions (up to now) addressed  in 
9232ea0

tests/test_cli.py
akorosov: The problem is that you may not have the `/tmp` folder, or not have writing permissions. `tempfile.TemporaryFile` creates temporary files in a safe and reliable manner. Please don't reinvent the wheel and hardcode paths. We should use the available Python functionality. It is the same as using `os.path.join(a,b)` instead of `a + '/' + b`. The latter seems simpler but it is not correct.

README.md
akorosov: The concept is exactly the same. `down_dir` is where you downloaded file will end up, right? That's what is called destination.
Please change the name.

tests/test_cli.py
akorosov: In these two tests you need some data for testing. The standard way of providing data in the database for testing is using fixtures.
Please move to fixtures.

akorosov (changes requested):
Please fix the use of temporary file, full name of the `-d` option, fixtures.

tests/test_cli.py
aanersc: These are two difference subjects! I did not reinvent the wheel! I have suggested to use `a.rstrip(os.path.sep) + os.path.sep + b.lstrip(os.path.sep)` not `a + '/' + b`. It is obvious that `a + '/' + b` is naive. I never suggested `a + '/' + b`.

`/tmp` folder is the suggestion of @aperrin66 and as far as I know, the linux machine always have that folder based on the https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard.

It seems I can't convince you two not to deal with a context manager!!!!
You are just making the tests complicated unnecessarily!
I will do it with context manager only because of your insist.

aanersc: @aperrin66 @akorosov 
I have just found that the address comes from the database and we need that file to be there based on that address which is installed to the database based on the fixtures! So it is impossible to user a tempfile as a context manager for this case. tempfile has a varying address but what is inside the fixtures is a constant address! Even `tempfile.NamedTemporaryFile` in addtion to `tempfile.TemporaryFile` does not provide a specific name for us to create a file with that name. THESE ARE ALL TEMPORARY.

Perhaps SOMETIMES it is better that you reconsider your decision(s) of your suggestions when I insist on something!

 Please review the code carefully, there are some time spend on writing the code, so it needs to spend time for reviewing it as well.

aanersc:
@aperrin66 
@akorosov 
All addressed in 4dbfe00 and 7791332. 7791332 is just a correction of what is wrongly pushed!

tests/test_cli.py
akorosov: If you 'have just found that' in the code that you wrote yourself, how do you expect me to find that? Please write good code that doesn't need to look up for hardcoded values in several places. Then it will be much easier to review (which is not an easy job either).

The problem with using `/tmp` is that it can potentially be unachievable. That's why `TemporaryFile` was developed and added to Python standard library - to provide safe and reliable access to temporary directories and files. As we have discussed many times, it is better to use standard approaches than develop something new. Usage of `os.path.join` is just an example. It is not related to usage of `/tmp`. It exemplifies that we can code either `a + '/' + 'b'` or `os.path.join(a, b)`. Although it seems more complicated, but the latter approach is correct because it uses the tools specially developed for this operation. The same here: we should not use '/tmp' because it is not a universal approach, it is not standard. It will fail, e.g. on Windows, or if you don't have access to '/tmp' or, maybe in other circumstances. Therefore we should not use it.

If we cannot use '/tmp' there are several ways to avoid using it. In some cases using `TemporaryFile` will help to avoid using '/tmp'. But apparently not in this case, because you hardcode this path in the fixtures. But it doesn't mean that we still should use '/tmp'. It means that we should look for other solutions. And that's what I expect from productive discussion of my review - attempt to identify the problem and attempt to find a solution to this problem. Not just a refusal of fixing problems.

If we cannot use `TemporaryFile` we can use something else. What's eventually needed in this test is to make a symbolic link in the call to `cli_copy.main()` and check that no errors are raised, right? And there are many ways to do that without using `/tmp`. Without even using `TemporaryDirectory`, or without actually creating a symbolic link.
Shall I suggest some?

akorosov (changes requested):

geospaas_processing/copiers.py
```diff
@@ -0,0 +1,72 @@
...
+    """Copier for datasets"""
+
+    def __init__(self, type_in_flag_file, destination_path, flag_file_request=False,
+                 link_request=False, **criteria):
+        self.TYPE_IN_FLAG_FILE = type_in_flag_file
+        self.DESTINATION_PATH = destination_path
+        self.FLAG_FILE_REQUEST = flag_file_request
+        self.LINK_REQUEST = link_request
+        self.DATASETS = Dataset.objects.filter(**criteria)
```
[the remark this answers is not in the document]
aanersc: Done in the next commit.

aanersc:
@akorosov 
`'have just found that'` is actually `'have just remembered that'`. It is because the time between the coding and receiving the review is large and I forget tiny aspects of details. Once again, I know that os.join is better than that plus sign between the strings. But, I never suggest or use `a + '/' + 'b'`.

I always welcome a productive discussion for coding but here in this case, I insist on this approach because it is the most silmilar case to the real case in production. The tests should be written in a way that they are most likely execute sth that is exactly the same as what happened in the real execution of the code. This could contains hardcoding approaches or other approaches. The approach does not matter. **The similarity between the test and what is happening in the real execution of the code** _**matters**_. Here in the case, the ways comes from database! So, at the previous versions I create a file and write the rest of the test and the @aperrin66 suggest to rename the folder to `/tmp`. Engaging the temporary file is not necessary unless we really need them! The subject of this test is something else! 

If the `/tmp` is problematic in the windows machines, then that a good reason for changing it. Then it maybe needed to user temporary files! MAYBE! So, what is your suggestion?

akorosov:
> If the `/tmp` is problematic in the windows machines, then that a good reason for changing it. Then it maybe needed to user temporary files! MAYBE! So, what is your suggestion?

Sure!
It is a unit test, right? We don't have to check functionality of the `os.symlink` function. We don't need to actually create a symlink (neither to check if it exists). Ideally unit tests should not make any changes to file system at all.
We just need to check that `os.symlink` was called with appropriate parameters. Simply mock it and check call arguments.

aanersc:
> > If the `/tmp` is problematic in the windows machines, then that a good reason for changing it. Then it maybe needed to user temporary files! MAYBE! So, what is your suggestion?
> 
> Sure!
> It is a unit test, right? We don't have to check functionality of the `os.symlink` function. We don't need to actually create a symlink (neither to check if it exists). Ideally unit tests should not make any changes to file system at all.
> We just need to check that `os.symlink` was called with appropriate parameters. Simply mock it and check call arguments.

No, what you are suggesting is done in this test with a multi-file manner of it with a crytal clear test:
https://github.com/nansencenter/django-geo-spaas-processing/blob/779133290386e238bfcdaf75ffb511dd42a68ad5/tests/test_cli.py#L278

By mocking the `os.symlink` we mock the actual duty of this package which is wrong! Mock is for other noisance things that nagging during our test. Here we test a copy action so we should not mock the copy action itself! This is the reason why this tests does not have a mock at the beginning.

The one possible solution that comes into my mind is that: 1. creating a namedTempfile in the test 2. save it in the database 3.call the function of copy 4 assert the result 5. delete that  namedTempfile
Are you agree with it?

aanersc:
@akorosov @aperrin66 
all addressed (lower_case_with_underscores and test removal)

akorosov (approved):
Fine with me. @aperrin66 , do you want to add something?

aperrin66 (approved):

Issue #30, opened 2021-01-07
deleter code


aanersc:

aanersc:
A code must be provide for deleting the OLD files after copying them

PR #31, opened 2021-01-08
#30 deleter_code_with_tests

aanersc:
@akorosov @aperrin66 
for the second new test, I mean for `test_deleting_symlinks_without_a_reference_file` it is not possible to work with temporary file. The reason is that the tempfile deals with the deletion of that tempfile inside the context manager in the __exit__ method. We have deleted the file in the middle of the test and this causes problem for that __exit__ method of tempfile to work properly. I use a normal file with `touch` instead. This second test is also in temporary directory, so it does not cause any problem and there is no need to change it in this regard.

### DIFF
--- a/geospaas_processing/cli/copy.py
+++ b/geospaas_processing/cli/copy.py
@@ -17,9 +17,8 @@ def main():
                                         flag_file_request=arg.flag_file,
                                         link_request=arg.link,
                                         destination_path=arg.destination_path,
-                                        time_to_live=int(arg.time_to_live),
-                                        keep_permanently=arg.keeping_permanently,
                                         **cumulative_query)
+    current_copy_action.delete(int(arg.time_to_live))
     current_copy_action.copy()
 
 
@@ -39,11 +38,6 @@ def cli_parse_args():
     parser.add_argument(
         '-t', '--type', required=False, type=str,
         help="The type of dataset (as a str) which is written in flag file for further processing.")
-    parser.add_argument(
-        '-k', '--keeping_permanently', required=False, action='store_true',
-        help="The flag that distinguishes between the two cases of 1.deletion of the previously "
-        + "copied file(s) after a certain period of time or 2.do nothing regarding deletion based "
-        + "on its ABSENCE or PRESENCE of this flag in the arguments.")
     parser.add_argument(
         '-ttl', '--time_to_live', required=False, type=str, default="90",
         help="The upper limit [in days] of file existence that have already been copied."
--- a/geospaas_processing/cli/copy.py
+++ b/geospaas_processing/cli/copy.py
@@ -17,11 +17,10 @@ def main():
                                         flag_file_request=arg.flag_file,
                                         link_request=arg.link,
                                         destination_path=arg.destination_path,
-                                        obsoleteness=int(arg.obsoleteness),
+                                        time_to_live=int(arg.time_to_live),
+                                        keep_permanently=arg.keeping_permanently,
                                         **cumulative_query)
     current_copy_action.copy()
-    if not arg.keeping_permanently:
-        current_copy_action.delete()
 
 
 def cli_parse_args():
@@ -46,8 +45,8 @@ def cli_parse_args():
         + "copied file(s) after a certain period of time or 2.do nothing regarding deletion based "
         + "on its ABSENCE or PRESENCE of this flag in the arguments.")
     parser.add_argument(
-        '-o', '--obsoleteness', required=False, type=str, default="90",
-        help="The upper limit of days of file existence that are being copied."
+        '-ttl', '--time_to_live', required=False, type=str, default="90",
+        help="The upper limit [in days] of file existence that have already been copied."
         + " If the file is older than this limit, it will be deleted.")
     return parser.parse_args()
 
--- a/geospaas_processing/cli/copy.py
+++ b/geospaas_processing/cli/copy.py
@@ -17,8 +17,11 @@ def main():
                                         flag_file_request=arg.flag_file,
                                         link_request=arg.link,
                                         destination_path=arg.destination_path,
+                                        obsoleteness=int(arg.obsoleteness),
                                         **cumulative_query)
     current_copy_action.copy()
+    if not arg.keeping_permanently:
+        current_copy_action.delete()
 
 
 def cli_parse_args():
@@ -37,6 +40,15 @@ def cli_parse_args():
     parser.add_argument(
         '-t', '--type', required=False, type=str,
         help="The type of dataset (as a str) which is written in flag file for further processing.")
+    parser.add_argument(
+        '-k', '--keeping_permanently', required=False, action='store_true',
+        help="The flag that distinguishes between the two cases of 1.deletion of the previously "
+        + "copied file(s) after a certain period of time or 2.do nothing regarding deletion based "
+        + "on its ABSENCE or PRESENCE of this flag in the arguments.")
+    parser.add_argument(
+        '-o', '--obsoleteness', required=False, type=str, default="90",
+        help="The upper limit of days of file existence that are being copied."
+        + " If the file is older than this limit, it will be deleted.")
     return parser.parse_args()
 
 
--- a/geospaas_processing/copiers.py
+++ b/geospaas_processing/copiers.py
@@ -17,14 +17,12 @@ LOGGER = logging.getLogger(__name__)
 class Copier():
     """Copier for datasets"""
 
-    def __init__(self, type_in_flag_file, destination_path, time_to_live, keep_permanently,
+    def __init__(self, type_in_flag_file, destination_path,
     flag_file_request=False, link_request=False, **criteria):
         self._type_in_flag_file = type_in_flag_file
         self._destination_path = destination_path
         self._flag_file_request = flag_file_request
         self._link_request = link_request
-        self.ttl = time_to_live
-        self.keep_permanently = keep_permanently
         self._datasets = Dataset.objects.filter(**criteria)
 
     @staticmethod
@@ -71,10 +69,7 @@ class Copier():
                     " there is no file in the stored address: %s.", dataset.id, source_path.uri)
 
     def copy(self):
-        """ First delete the old files then tries to copy all datasets based on their stored local
-        addresses in the database."""
-        if not self.keep_permanently:
-            self.delete()
+        """ Tries to copy all datasets based on their stored local addresses in the database."""
         for dataset in self._datasets:
             if dataset.dataseturi_set.filter(service=LOCAL_FILE_SERVICE).exists():
                 source_paths = dataset.dataseturi_set.filter(service=LOCAL_FILE_SERVICE)
@@ -83,7 +78,7 @@ class Copier():
                 LOGGER.debug("For dataset with id = %s, there is no local file address in the "
                              "database.", dataset.id)
 
-    def delete(self):
+    def delete(self, ttl):
         """
         Delete the file(s) or symlink(s) after a certain period of 'time to live' (in days) of the
         file(s) or symlink(s) inside the destination path. """
@@ -92,5 +87,5 @@ class Copier():
                 if ((entry.is_file(follow_symlinks=False) or entry.is_symlink())
                     and '.snapshot' not in entry.path
                     and entry.stat(follow_symlinks=False).st_uid == os.getuid()
-                    and time.time() - entry.stat(follow_symlinks=False).st_mtime > self.ttl*24*3600):
+                    and time.time() - entry.stat(follow_symlinks=False).st_mtime > ttl*24*3600):
                         os.remove(entry.path)
--- a/geospaas_processing/copiers.py
+++ b/geospaas_processing/copiers.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import logging
+import time
 
 import django
 
@@ -16,12 +17,13 @@ LOGGER = logging.getLogger(__name__)
 class Copier():
     """Copier for datasets"""
 
-    def __init__(self, type_in_flag_file, destination_path, flag_file_request=False,
+    def __init__(self, type_in_flag_file, destination_path, obsoleteness, flag_file_request=False,
                  link_request=False, **criteria):
         self._type_in_flag_file = type_in_flag_file
         self._destination_path = destination_path
         self._flag_file_request = flag_file_request
         self._link_request = link_request
+        self.obsoleteness = obsoleteness
         self._datasets = Dataset.objects.filter(**criteria)
 
     @staticmethod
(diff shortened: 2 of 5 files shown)


aanersc:
This is the inside part of the code which needs a version after review. The outside part of the code will be pushed in the adas deploy repo.
@akorosov @aperrin66

geospaas_processing/cli/copy.py
```diff
@@ -37,6 +40,15 @@ def cli_parse_args():
     parser.add_argument(
         '-t', '--type', required=False, type=str,
         help="The type of dataset (as a str) which is written in flag file for further processing.")
+    parser.add_argument(
+        '-k', '--keeping_permanently', required=False, action='store_true',
```
aperrin66, line 44: The long name of the argument is a bit confusing. There is no guarantee that the files will be kept permanently.

geospaas_processing/copiers.py
```diff
@@ -16,12 +17,13 @@
...
     """Copier for datasets"""
 
-    def __init__(self, type_in_flag_file, destination_path, flag_file_request=False,
+    def __init__(self, type_in_flag_file, destination_path, obsoleteness, flag_file_request=False,
                  link_request=False, **criteria):
         self._type_in_flag_file = type_in_flag_file
         self._destination_path = destination_path
         self._flag_file_request = flag_file_request
         self._link_request = link_request
+        self.obsoleteness = obsoleteness
```
aperrin66, line 26: Why not just put this "obsoleteness" as a parameter to the `delete()` method?

geospaas_processing/copiers.py
```diff
@@ -76,3 +78,18 @@ def copy(self):
...
+    def delete(self):
+        """
+        Delete the file(s) or symlink(s) after a certain period of obsoleteness (in days) of the
+        file(s) or symlink(s) inside the destination path. """
+        with os.scandir(self._destination_path) as scanned_dir:
+            for entry in scanned_dir:
+                if ('.snapshot' not in entry.path
+                and entry.stat(follow_symlinks=False).st_uid == os.getuid()
+                and time.time() - entry.stat(follow_symlinks=False).st_mtime >
+                    self.obsoleteness*24*3600):
```
aperrin66, line 91: Please fix the indentation.

geospaas_processing/cli/copy.py
aperrin66: Maybe this parameter is not necessary at all, you could probably make it work with the "obsoleteness" parameter alone.

geospaas_processing/copiers.py
```diff
@@ -76,3 +78,18 @@ def copy(self):
...
+                    if (entry.is_file(follow_symlinks=False)):
+                        os.remove(entry.path)
+                    elif entry.is_symlink():
+                        os.unlink(entry.path)
```
aperrin66, line 95: This code is equivalent to:
```python
if entry.is_file():
    os.remove(entry.path)
```

aperrin66 (changes requested):

geospaas_processing/cli/copy.py
aanersc: Yeah, that is the way I think first about it. We can set it to zero to dsable this ability of deletion. But @akorosov disagreed and requested for another cli input named `keep_permanently`. so both of us, please tell me to write the code with which approach? working with zero? or having another input like `keep_permanently`?

geospaas_processing/cli/copy.py
```diff
@@ -37,6 +40,15 @@ def cli_parse_args():
...
+        help="The flag that distinguishes between the two cases of 1.deletion of the previously "
+        + "copied file(s) after a certain period of time or 2.do nothing regarding deletion based "
+        + "on its ABSENCE or PRESENCE of this flag in the arguments.")
+    parser.add_argument(
+        '-o', '--obsoleteness', required=False, type=str, default="90",
```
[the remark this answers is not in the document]
aanersc: Do you want me to change the name from `obsoleteness `to `ttl`?

geospaas_processing/cli/copy.py
```diff
@@ -37,6 +40,15 @@ def cli_parse_args():
...
+        help="The upper limit of days of file existence that are being copied."
```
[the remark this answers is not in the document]
aanersc: ok, then `that are being copied`should be replaced with `that have already been copied`.right?

geospaas_processing/copiers.py
aanersc: Because of OOP paradigm, the obsoleteness is sth that is related to the object(it belongs to the object), not a parameters that can be carried and varied between different functions of an object. SO it is needed to be defined as an attribute of the very object. It will be easier to debug in the future if necessary.

geospaas_processing/copiers.py
aanersc: How? Do you mean removing it? I dont know that you mean exactly.

geospaas_processing/copiers.py
aanersc: In the case of having a healthy symlink, `os.remove` removes the original file! which is not what we want here in this code. We want that symlink which is in the spool folder is being deleted! not the original file. That is the reason why I differentiate between two case of `having symlink file in the spool folder` and `having a file in the spool folder`.

aperrin66: `os.remove()` and `os.unlink()` behave the same regarding symlinks.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aperrin66: I just gave you some information, do what you want with it.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aperrin66: Yes, and also specify the unit.

geospaas_processing/copiers.py
aperrin66: Add 4 or 8 spaces in the beginning of the lines which begin with `and`.

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: ttl is a better choice I guess

geospaas_processing/cli/copy.py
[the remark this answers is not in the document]
aanersc: Done in next commit

geospaas_processing/cli/copy.py
```diff
@@ -17,8 +17,11 @@ def main():
                                         flag_file_request=arg.flag_file,
                                         link_request=arg.link,
                                         destination_path=arg.destination_path,
+                                        obsoleteness=int(arg.obsoleteness),
                                         **cumulative_query)
     current_copy_action.copy()
+    if not arg.keeping_permanently:
+        current_copy_action.delete()
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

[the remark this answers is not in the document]
aanersc: Done in the next commit

geospaas_processing/cli/copy.py
aanersc: discussed

geospaas_processing/copiers.py
aanersc: Done in the next commit

geospaas_processing/copiers.py
aanersc: Done in the next commit

aanersc:
@aperrin66  all addressed

geospaas_processing/cli/copy.py
aperrin66: The part about the name is still valid though.

geospaas_processing/copiers.py
```diff
@@ -68,11 +71,26 @@ def file_or_symlink_copy(self, source_paths, dataset):
                     " there is no file in the stored address: %s.", dataset.id, source_path.uri)
 
     def copy(self):
-        """ Tries to copy all datasets based on their stored local addresses in the database."""
+        """ First delete the old files then tries to copy all datasets based on their stored local
+        addresses in the database."""
+        if not self.keep_permanently:
+            self.delete()
```
akorosov, line 77: Move call to delete to script

geospaas_processing/copiers.py
```diff
@@ -68,11 +71,26 @@ def file_or_symlink_copy(self, source_paths, dataset):
                     " there is no file in the stored address: %s.", dataset.id, source_path.uri)
 
     def copy(self):
-        """ Tries to copy all datasets based on their stored local addresses in the database."""
+        """ First delete the old files then tries to copy all datasets based on their stored local
+        addresses in the database."""
+        if not self.keep_permanently:
```
akorosov, line 76: remove keep_permanently

geospaas_processing/cli/copy.py
```diff
@@ -17,6 +17,8 @@ def main():
                                         flag_file_request=arg.flag_file,
                                         link_request=arg.link,
                                         destination_path=arg.destination_path,
+                                        time_to_live=int(arg.time_to_live),
+                                        keep_permanently=arg.keeping_permanently,
                                         **cumulative_query)
     current_copy_action.copy()
```
akorosov, line 23: Add `current_copy_action.delete(int(arg.time_to_live))` before call to `copy()`

akorosov (changes requested):

aanersc:
@aperrin66 @akorosov addressed

geospaas_processing/copiers.py
```diff
@@ -76,3 +77,15 @@ def copy(self):
...
+        """
+        Delete the file(s) or symlink(s) after a certain period of 'time to live' (in days) of the
+        file(s) or symlink(s) inside the destination path. """
+        with os.scandir(self._destination_path) as scanned_dir:
+            for entry in scanned_dir:
+                if ((entry.is_file(follow_symlinks=False) or entry.is_symlink())
+                    and '.snapshot' not in entry.path
+                    and entry.stat(follow_symlinks=False).st_uid == os.getuid()
+                    and time.time() - entry.stat(follow_symlinks=False).st_mtime > ttl*24*3600):
+                        os.remove(entry.path)
```
akorosov, line 91:
```suggestion
                    os.remove(entry.path)
```

tests/test_cli.py
```diff
@@ -293,6 +299,48 @@ def test_correct_place_of_symlink_after_creation_of_it(self, mock_isfile):
...
+                sys.argv = [
+                    "",
+                    '-b', "2038-06-01",
+                    '-e', "2038-06-09",
+                    '-d', tmpdirname,
+                    "--time_to_live", "0"
+                ]
+                with mock.patch('shutil.copy'):
+                    with mock.patch('os.remove') as mock_os_remove:
+                        cli_copy.main()
```
akorosov, line 317: Reduce your untitests to bare minimum:
```suggestion
                copier = Copier('type1', tmpdirname)
                copier.delete(0)
```

tests/test_cli.py
```diff
@@ -293,6 +299,48 @@ def test_correct_place_of_symlink_after_creation_of_it(self, mock_isfile):
...
+    def test_deleting_symlinks_without_a_reference_file(self, mock_copy):
+        """ delete function should delete the symlink(s) regardless of the state of existence of the
+        reference file. In this test, the symlink is existing without a reference file. The test
+        shall remove that symlink."""
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            path_of_file = os.path.join(tmpdirname, 'test_file_name')
+            Path(path_of_file).touch()
+            symlink_path = path_of_file + '_symlink'
+            os.symlink(src=path_of_file, dst=symlink_path)  # symlink creation
+            os.remove(path_of_file)  # remove the reference file of the symlink
```
akorosov, line 332: Move that to the test above. Let this one test have three cases: one file, one valid symlink, one invalid symlink.

akorosov (changes requested):
Simplify testing: remove call to cli_copy.main(), merge two tests.

geospaas_processing/copiers.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: I have done that, the mocks become confusing and the tests won't run correctly, it is better to test them separately:
image

image

tests/test_cli.py
aanersc: done in the next commit

aanersc:
@akorosov all addressed

akorosov (changes requested):
Simplify testing: 
- [ ] merge two tests.

tests/test_cli.py
```diff
@@ -318,4 +356,4 @@ def test_correct_content_of_flag_file(self, mock_isfile, mock_link):
                 "05T00:43:05.826008Z, Instrument: SLSTR, Mode: EO, Satellite: Sentinel-3, Size: 415"
                 ".29 MB\n"
             )
-        )
+            )
```
aperrin66, line 351: Since both parentheses are opened on the same line, please close them on the same line

aanersc:
@akorosov @aperrin66 your code is pushed.

tests/test_cli.py
```diff
@@ -293,6 +299,30 @@ def test_correct_place_of_symlink_after_creation_of_it(self, mock_isfile):
...
+        test a temporary file and a symlink of it is created. Both of them should be deleted.
+        delete function should also delete the symlink(s) regardless of the state of existence of
+        the reference file. In this test, the symlink is existing without a reference file. The test
+        shall remove that symlink.
+        """
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            with tempfile.NamedTemporaryFile(dir=tmpdirname) as tmpfile:
+                os.symlink(src=tmpfile.name, dst=tmpfile.name + '_symlink')  # symlink creation
+                broken_symlink_path = os.path.join(tmpdirname, 'broken_symlink')
+                os.symlink(src='/tmp/blah', dst=broken_symlink_path)
```
aperrin66, line 313: The code I sent you was just an example, please use a source path inside your temporary directory, so that you can be sure it's not an existing file.

aperrin66 (changes requested):
I pointed out a few things in inline comments.

While you are at it, there is an indentation error on line 60 of the `copiers.py` file (one space is missing). Can you please fix that?

On a general note, please be careful about indentation.
It might seem unimportant, but it's a huge readability factor.

geospaas_processing/copiers.py
```diff
@@ -16,8 +17,8 @@
 class Copier():
     """Copier for datasets"""
 
-    def __init__(self, type_in_flag_file, destination_path, flag_file_request=False,
-                 link_request=False, **criteria):
+    def __init__(self, type_in_flag_file, destination_path,
+    flag_file_request=False, link_request=False, **criteria):
```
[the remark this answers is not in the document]
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

tests/test_cli.py
aanersc: Done in the next commit

aanersc:
@aperrin66 all addressed

akorosov (approved):

aperrin66 (approved):

PR #33, opened 2021-01-14
Update delete_and_copy.py

aanersc:
@akorosov @aperrin66 based on the new design of lacking the --keep_permanently, ttl must be provided as input anytime when using delete_and_copy cli


aanersc:
@akorosov @aperrin66 I must work more on it!! It is not done with single line modification!!!!

aperrin66:
Is this really necessary? There is a default value for the parameter.

aanersc:
> Is this really necessary? There is a default value for the parameter.

yeah,you are right! NO no to do anything in this regard.

Issue #34, opened 2021-01-20
Copy code should copy the whole folder


aanersc:
Up to now, there is no code for foldernames because crawler does not return a folder name. Based on issue https://github.com/nansencenter/django-geo-spaas-harvesting/pull/68#issue-557368094 the copy code should also copy the whole folder because (for local harvester), sometimes, the crawler returns the whole folder if it reaches a folder name.

PR #35, opened 2021-01-22
Issue#34 copier of folders

aanersc:
@aperrin66 @akorosov 
After developing crawler preventer, as mentioned in #34 we lack a code for copiers of folders! Here it is! I have also updated the version number with this PR.

### DIFF
--- a/geospaas_processing/copiers.py
+++ b/geospaas_processing/copiers.py
@@ -2,8 +2,7 @@ import os
 import shutil
 import logging
 import time
-from distutils.dir_util import copy_tree
-
+from os.path import exists
 import django
 
 from geospaas.catalog.models import Dataset
@@ -46,21 +45,17 @@ class Copier():
         """copy the file or a symlink of the file/folder of dataset based on its stored local
         address in the database."""
         for source_path in source_paths:
-            self.source_is_file = os.path.isfile(source_path.uri)
-            self.source_is_folder = os.path.isdir(source_path.uri)
             destination_file_or_folder_name = os.path.join(
                 self._destination_path, os.path.basename(source_path.uri))
-            if (self.source_is_file or self.source_is_folder):
+            if exists(source_path.uri):
                 # below if condition prevents "shutil.copy" or "os.symlink" from replacing the file
                 # or folder in the destination in a repetitive manner.
-                if (not (os.path.isfile(destination_file_or_folder_name)
-                         or os.path.isdir(destination_file_or_folder_name))
-                    or not os.path.islink(destination_file_or_folder_name)):
+                if not exists(destination_file_or_folder_name):
                     self.copy_item(source_path, destination_file_or_folder_name, dataset)
                 else:
-                    LOGGER.debug(
-                        "For dataset with id = %s, there is already a symlink or a file or a folder"
-                        + "with the same name in the destination folder.", dataset.id)
+                    LOGGER.warning(
+                        "Failed to copy dataset %s: the destination path already exists.",
+                        dataset.id)
             else:
                 LOGGER.debug(
                     "For stored address of dataset with id = %s,"
@@ -73,10 +68,10 @@ class Copier():
         if self._link_request:
             os.symlink(src=source_path.uri, dst=destination_file_or_folder_name)
         else:
-            if self.source_is_file:
+            if os.path.isfile(source_path.uri):
                 shutil.copy(src=source_path.uri, dst=self._destination_path)
-            elif self.source_is_folder:
-                copy_tree(src=source_path.uri, dst=os.path.join(
+            elif os.path.isdir(source_path.uri):
+                shutil.copytree(src=source_path.uri, dst=os.path.join(
                     self._destination_path, os.path.basename(source_path.uri)))
         if self._flag_file_request:
             self.write_flag_file(self._type_in_flag_file,
@@ -100,7 +95,7 @@ class Copier():
         with os.scandir(self._destination_path) as scanned_dir:
             for entry in scanned_dir:
                 if ((entry.is_file(follow_symlinks=False) or entry.is_symlink())
-                    and '.snapshot' not in entry.path
-                    and entry.stat(follow_symlinks=False).st_uid == os.getuid()
+                        and '.snapshot' not in entry.path
+                        and entry.stat(follow_symlinks=False).st_uid == os.getuid()
                         and time.time() - entry.stat(follow_symlinks=False).st_mtime > ttl*24*3600):
                     os.remove(entry.path)
--- a/geospaas_processing/copiers.py
+++ b/geospaas_processing/copiers.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import logging
 import time
+from distutils.dir_util import copy_tree
 
 import django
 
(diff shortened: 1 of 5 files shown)


geospaas_processing/copiers.py
```diff
@@ -2,6 +2,7 @@
 import shutil
 import logging
 import time
+from distutils.dir_util import copy_tree
```
aperrin66, line 5: There is already a `copytree()` function in `shutil`, no need for an extra import.

geospaas_processing/copiers.py
```diff
@@ -42,31 +43,44 @@ def write_flag_file(type_in_flag_file, source_path, dataset, destination_filenam
...
-        """copy the file or a symlink of the file of dataset based on its stored local address in
-        the database."""
+        """copy the file or a symlink of the file/folder of dataset based on its stored local
+        address in the database."""
         for source_path in source_paths:
-            if os.path.isfile(source_path.uri):
-                destination_filename = os.path.join(
-                    self._destination_path, os.path.basename(source_path.uri))
+            self.source_is_file = os.path.isfile(source_path.uri)
+            self.source_is_folder = os.path.isdir(source_path.uri)
```
aperrin66, line 50: Please don't use object attributes like pseudo global variables.
For this usage, it's much safer to pass the information as arguments to the `copy_item()` method, because if `copy_item()` is called from somewhere else than `file_or_symlink_copy()`, it will rely on the wrong information (`self.source_is_file()` and `self.source_is_folder()` will have the values of the last call to `file_or_symlink_copy()`).

Alternatively, just use `os.path.exists()` in `file_or_symlink_copy()`, and call `os.path.isfile()` and `os.path.isdir()` directly in `copy_item()`

geospaas_processing/copiers.py
```diff
@@ -42,31 +43,44 @@ def write_flag_file(type_in_flag_file, source_path, dataset, destination_filenam
...
+                if (not (os.path.isfile(destination_file_or_folder_name)
+                         or os.path.isdir(destination_file_or_folder_name))
+                    or not os.path.islink(destination_file_or_folder_name)):
+                    self.copy_item(source_path, destination_file_or_folder_name, dataset)
                 else:
                     LOGGER.debug(
-                        "For dataset with id = %s, there is already a symlink or a file with the "
-                        + "same name in the destination folder.", dataset.id)
+                        "For dataset with id = %s, there is already a symlink or a file or a folder"
+                        + "with the same name in the destination folder.", dataset.id)
```
aperrin66, line 63: That should be at least a warning, if not an error message.

aperrin66: It can be made shorter too, something like: "Failed to copy dataset %s: the destination path already exists."

geospaas_processing/copiers.py
```diff
@@ -88,5 +102,5 @@ def delete(self, ttl):
                 if ((entry.is_file(follow_symlinks=False) or entry.is_symlink())
                     and '.snapshot' not in entry.path
                     and entry.stat(follow_symlinks=False).st_uid == os.getuid()
-                    and time.time() - entry.stat(follow_symlinks=False).st_mtime > ttl*24*3600):
+                        and time.time() - entry.stat(follow_symlinks=False).st_mtime > ttl*24*3600):
```
aperrin66, line 105: If you do that, please also add an indentation level in front of the other conditions (beginning with `and`).

aperrin66 (changes requested):

geospaas_processing/copiers.py
aanersc: Done

geospaas_processing/copiers.py
aanersc: Done in the next commit.

geospaas_processing/copiers.py
aanersc: Done

geospaas_processing/copiers.py
aanersc: NO,I have not done that. It is from autopep8

geospaas_processing/copiers.py
aanersc: Done

geospaas_processing/copiers.py
aanersc: Done

aanersc:
@aperrin66 ready

geospaas_processing/copiers.py
```diff
@@ -42,31 +42,40 @@ def write_flag_file(type_in_flag_file, source_path, dataset, destination_filenam
...
-        the database."""
+        """copy the file or a symlink of the file/folder of dataset based on its stored local
+        address in the database."""
         for source_path in source_paths:
-            if os.path.isfile(source_path.uri):
-                destination_filename = os.path.join(
-                    self._destination_path, os.path.basename(source_path.uri))
+            destination_file_or_folder_name = os.path.join(
+                self._destination_path, os.path.basename(source_path.uri))
+            if exists(source_path.uri):
```
aperrin66, line 50: I think it would be clearer to import `os.path` and use `os.path.exists()` explicitly, because "exists" is a very general term which can be applied to a lot of different subjects.

aanersc: That way we can not test it. Because .exists() is used .isfile() and .isdir() functions. And it is impossible to patch it in the tests. This way is more clearer in both code and test files

aanersc: if you want, I can use an alias such as `path_exists `instead of `exists `?!?

aperrin66: No, it does not prevent you from mocking.
We already discussed this at length in the first [copy CLI pull request](https://github.com/nansencenter/django-geo-spaas-processing/pull/24), about mocking `os.path.isfile`.
Please use the same solution.

aanersc: Maybe I convey it wrong. Here this is not the case (the previous case). Look at this comment in that PR: https://github.com/nansencenter/django-geo-spaas-processing/pull/24#discussion_r527683675. Those are independant. But here, .isfile() is dependant to .exists(). So if I patch .exists() then isfile is not working properly.

aperrin66: Ok, I did some testing and patching `os.path.exists()` causes a problem, but not for the reason you think.
`isfile()` does not use `exists()`, it uses `os.stat()`.
But here, what causes a problem is that `os.path.exists()` is used by the argument parser in the CLI.

So let's just leave it like that.

aperrin66: However, this raises the point that for unit tests, you don't need to use the argument parsing in every test.
You need to have tests which test that the copier is correctly called when the CLI is given some arguments, and for the rest you can just test the copying functionalities by caling the copier methods directly.

The word "unit" in "unit tests" means that you test the smallest unit of code at a time.

geospaas_processing/copiers.py
```diff
@@ -42,31 +42,40 @@ def write_flag_file(type_in_flag_file, source_path, dataset, destination_filenam
...
-                        + "same name in the destination folder.", dataset.id)
+                    LOGGER.warning(
+                        "Failed to copy dataset %s: the destination path already exists.",
+                        dataset.id)
             else:
                 LOGGER.debug(
                     "For stored address of dataset with id = %s,"
-                    " there is no file in the stored address: %s.", dataset.id, source_path.uri)
+                    " there is no file or no folder in the stored address: %s.", dataset.id,
+                    source_path.uri)
```
aperrin66, line 63: https://github.com/nansencenter/django-geo-spaas-processing/pull/35#discussion_r562500486
Same comment here.

aperrin66 (changes requested):
The error cases are not tested (no source or destination file, etc...).
See https://coveralls.io/builds/36523884/source?filename=%2Fsrc%2Fgeospaas_processing%2Fcopiers.py

geospaas_processing/copiers.py
aanersc: Done

geospaas_processing/copiers.py
```diff
@@ -2,7 +2,7 @@
 import shutil
 import logging
 import time
-
+from os.path import exists
```
akorosov, line 5: You use that only twice. Please use it as `os.path.exists`
```suggestion
```

tests/test_cli.py
```diff
@@ -258,11 +262,13 @@ def test_lack_of_calling_json_deserializer_when_no_query_appears_for_copying(sel
         mock_json.assert_not_called()
 
     @mock.patch('geospaas_processing.copiers.Copier.delete')
-    @mock.patch('os.path.isfile', side_effect=[True, False, True, False])
-    # The even side effects (the 'True' ones) are associated to the destination and the odd ones are
-    # associated to the source path. It is because 'os.path.isfile' is used for evaluating both
-    # source paths and destination paths.
-    def test_correct_destination_folder_for_all_files_that_are_copied(self, mock_isfile, mock_del):
+    @mock.patch('os.path.isfile', return_value=True)
+    @mock.patch('geospaas_processing.copiers.exists', side_effect=[True, False, True, False])
```
akorosov, line 266:
```suggestion
    @mock.patch('geospaas_processing.copiers.os.path.exists', side_effect=[True, False, True, False])
```

tests/test_cli.py
```diff
@@ -274,14 +280,46 @@ def test_correct_destination_folder_for_all_files_that_are_copied(self, mock_isf
...
-        self.assertEqual(
-            [call(dst='/dst_folder/', src='/tmp/testing_file.test'),
+        self.assertCountEqual(
+            [call(dst='/dst_folder/', src='/tmp/testing_file_or_folder.test'),
              call(dst='/dst_folder/', src='/new_loc_add')],
             mock_copy.call_args_list)
 
+    @mock.patch('geospaas_processing.copiers.Copier.delete')
+    @mock.patch('os.path.isdir', return_value=True)
+    @mock.patch('geospaas_processing.copiers.exists', side_effect=[True, False, True, False])
```
akorosov, line 290:
```suggestion
    @mock.patch('geospaas_processing.copiers.os.path.exists', side_effect=[True, False, True, False])
```

akorosov (approved):
Minor thing regarding import os. Otherwise seems ok.

geospaas_processing/copiers.py
aanersc: @akorosov it was discussed here: https://github.com/nansencenter/django-geo-spaas-processing/pull/35#discussion_r562654203

We can't do that!

tests/test_cli.py
aanersc: @akorosov it was discussed here: https://github.com/nansencenter/django-geo-spaas-processing/pull/35#discussion_r562654203

We can't do that!

tests/test_cli.py
aanersc: same as above

aperrin66 (approved):

PR #39, opened 2021-02-23
FTP downloader

### LINKED ISSUE #26: Make a FTP downloader

aperrin66:
Resolves #26 

I took the opportunity to:
- move common functionalities to the base Downloader class
- remove the file prefixes

### DIFF
--- a/geospaas_processing/downloaders.py
+++ b/geospaas_processing/downloaders.py
@@ -8,9 +8,11 @@ The Redis instance hostname and port can be set via the following environment va
   - GEOSPAAS_PROCESSING_REDIS_PORT
 """
 import errno
+import ftplib
 import logging
 import os
 import os.path
+from urllib.parse import urlparse
 
 import requests
 import requests.utils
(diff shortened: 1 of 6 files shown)


geospaas_processing/downloaders.py
```diff
@@ -39,27 +41,104 @@ class TooManyDownloadsError(DownloadError):
...
+        """
+        raise NotImplementedError()
+
+    @classmethod
+    def close_connection(cls, connection):
+        """Closes the open connection. Most objects used as connections
+        have a close() method, but it might be necessary to override
+        this in child classes.
+        """
+        connection.close()
```
aanersc, line 67: Is there any `close()` method for the response of `requests`library? I am asking this because you return it in the case HTTP. How you test the downling in the HTTP case with this PR?

geospaas_processing/downloaders.py
```diff
@@ -71,94 +150,84 @@ def extract_file_name_from_response(response):
...
-
-        # Try to free some space if we can get the size of the file about to be downloaded
-        file_size = cls.get_remote_file_size(response, auth)
-        if file_size:
-            LOGGER.debug("Checking there is enough free space to download %s bytes", file_size)
-            utils.LocalStorage(path=download_dir).free_space(file_size)
+    @classmethod
+    def get_file_name(cls, url, connection):
+        """Extracts the file name from the URL"""
+        return urlparse(url).path.split('/')[-1] or None
```
aanersc, line 216: use os.path.basename and dirname if needed instead of`split('/')[-1]`

https://docs.python.org/3/library/os.path.html#os.path.dirname

tests/test_cli.py
```diff
@@ -92,55 +88,6 @@ def test_lack_of_calling_json_deserializer_when_no_query_appears(
...
-            GEOSGeometry('POLYGON ((-22 84, -22 74, 32 74, 32 84, -22 84))'),
-            'max_downloads': 100,
-            'provider_settings_path': '/config_folder/config_file.yml',
-            'time_coverage_end__lte': datetime(2020, 8, 22, 0, 0, tzinfo=tzutc()),
-            'time_coverage_start__gte': datetime(2019, 10, 22, 0, 0, tzinfo=tzutc()),
-            'dataseturi__uri__contains': 'osisaf',
-            'source__instrument__short_name__icontains': 'AMSR2',
-            'use_file_prefix': True,
-            'save_path': True
-        }, mock_download_manager_init.call_args)
```
aanersc, line 142: this test is the most comprehensive one compared to the one next to it and compared to the previous one. So keep this one and delete the next and the previous one entirely. Because this one has the geometry in it and by deleting that we do not test the geometry input.

aanersc (commented):
noted inline

geospaas_processing/downloaders.py
aperrin66: I am dealing with a URL, not a path, so this is not OS dependent.
By using `os.path` I would have problems on OS which do not use '/' as a path separator.

geospaas_processing/downloaders.py
aperrin66: Yes, `requests.Response` has a close method and there are unit tests for HTTP downloads.

tests/test_cli.py
aperrin66: Isn't the geometry input tested in `test_correct_call_DownloadManager_with_relative_time()`?

aanersc: yes.no need to change

akorosov (approved):

PR #46, opened 2021-03-25
Add local downloader

### LINKED ISSUE #45: Enable the "download" of locally hosted files
Create a downloader which takes care of copying locally hosted files to a working directory.

aperrin66:
Resolves #45

### DIFF
--- a/geospaas_processing/downloaders.py
+++ b/geospaas_processing/downloaders.py
@@ -12,6 +12,7 @@ import ftplib
 import logging
 import os
 import os.path
+import shutil
 from urllib.parse import urlparse
 
 import requests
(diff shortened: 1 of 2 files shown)


geospaas_processing/downloaders.py
```diff
@@ -230,6 +231,36 @@ def download_file(cls, file, url, connection):
...
+        return os.path.basename(url)
+
+    @classmethod
+    def get_file_size(cls, url, connection, auth=(None, None)):
+        return os.path.getsize(url)
+
+    @classmethod
+    def download_file(cls, file, url, connection):
+        with open(url, 'rb') as source:
+            shutil.copyfileobj(source, file)
```
azamifarda, line 261: why do you define this methods as classmethods? shouldn't be as a normal object method?

azamifarda (commented):
noted inline

geospaas_processing/downloaders.py
aperrin66: Because they are defined like that in the parent class.
And they are like that in the parent class because Downloaders don't use any instance attributes.

azamifarda (approved):

akorosov (approved):

PR #56, opened 2021-04-22
Fix scihub concurrent download

aperrin66:
- Use a HEAD request to get the file name for the HTTP downloader.
  This avoids problems with the stream connection not closing when a file already exists.
- Make sure that connections are always closed.
- Make error messages clearer.


aperrin66:
The build is not passing because coveralls.io is down.

akorosov (approved):

PR #58, opened 2021-05-03
Deal with Scihub's API URL change

### LINKED ISSUE #57: Copernicus apihub URL changed
https://scihub.copernicus.eu/news/News00868

The URL changed from
https://scihub.copernicus.eu/apihub/
to
https://apihub.copernicus.eu/apihub/

There is a redirection, but `requests` removes authentication headers when a redirection to a different host occurs:
https://docs.python-requests.org/en/master/user/quickstart/#custom-headers

aperrin66:
Resolves #57

### DIFF
--- a/geospaas_processing/utils.py
+++ b/geospaas_processing/utils.py
@@ -10,8 +10,10 @@ import time
 import yaml
 import zipfile
 from contextlib import contextmanager
+from urllib.parse import urlparse
 
 import paramiko
+import requests
 import scp
 try:
     from redis import Redis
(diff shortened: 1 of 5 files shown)


geospaas_processing/utils.py
```diff
@@ -330,3 +332,37 @@ def yaml_env_safe_load(stream):
...
+    inside a TrustDomainSession if authentication is provided. This
+    makes it possible to follow redirections inside the same domain.
+    """
+    auth = kwargs.pop('auth', None)
+    if auth:
+        with TrustDomainSession() as session:
+            session.auth = auth
+            return session.request(http_method, *args, **kwargs)
+    else:
+        return requests.request(http_method, *args, **kwargs)
```
azamifarda, line 368: We have duplication of code here. Can you place the code in the utils of geospaas so that both harvesting and processing use only one single code?

azamifarda (changes requested):

geospaas_processing/utils.py
aperrin66: `geospaas_harvesting` and `geospaas_processing` are different packages, so the only way to do that would be to put the utilities in a separate package, which both harvesting and processing would have as a dependency. This means a new repository and CI, version management, etc.
I don't think it's worth it while the common code is that small.

azamifarda: No, I mean put the code in the geospaas repo, not another repo. Obviously another repo is not required here for this tiny code.

aperrin66: Do you mean putting it in https://github.com/nansencenter/django-geo-spaas/blob/master/geospaas/utils/utils.py ?

azamifarda: yes

aperrin66: Hum, I don't know what the best solution would be.
Any thoughts on this @akorosov?

akorosov: Let's make a new issue in geospaas for that.

akorosov (approved):

akorosov:
@aperrin66 , can you rebase and merge, please?

PR #70, opened 2022-03-09
Download datasets in a unique path

aperrin66:
This is necessary because for some datasets, the file doesn't contain the date, so caching a file prevents from downloading files with the same name in the future.
Datasets are now downloaded in a folder named using the entry_id of the dataset, ensuring uniqueness.


aperrin66:
FYI @akorosov

PR #74, opened 2023-02-21
Add Syntool preprocessing capabilities

aperrin66:
Add the ability to convert downloaded dataset files to the format needed to display it in Syntool.
Also added the possibility to trigger harvesting via a task.
At the same time, I altered a bit the structure to split processing tasks with different themes/scopes in different modules:
- `tasks.core`: operations that can be applied to all datasets like download or archive.
- `tasks.idf`: IDF conversion
- `tasks.syntool`: Syntool conversion
- `tasks.harvesting`: harvesting

#73 was also merged in here, adding the ability to crop dataset files to a given bounding box.


aperrin66:
Discussed with @akorosov , merging.

